PR #9, opened 2020-11-10
Restoring old input values after a failed validation

wojciechgabrys:
When redirecting the user back after a failed validation (while attempting to register, sign in, or resetting the password), the `name` and `email` fields are now filled with the old values.


driesvints:
@wojciechgabrys `{{ }}` syntax isn't supported for Blade components. I've sent in a fix for this here: https://github.com/laravel/breeze/pull/15

PR #10, opened 2020-11-11
install alpinejs via npm instead of cdn import

ahmed-aliraqi:
This PR. allow to use `alpinejs` locally from `npm` instead of import CDN link.


thewebartisan7:
bootstrap.js is missing

driesvints:
@thewebartisan7 it's in your laravel installation already: https://github.com/laravel/laravel/blob/8.x/resources/js/bootstrap.js

PR #7, opened 2020-11-10
Added new logo for Laravel Breeze

mnaderian:
Created a new logo for Laravel Breeze. Hope you like it.

PR #12, opened 2020-11-11
[1.x] Logo for Breeze.

mnaderian:
Updated the logo from #7 
Changed it to be just about the same as requested. Hope you like it.

Laravel Breeze Logo

**PS:** _Sorry, I don't wanna impose my design. If you don't like this kind of logo that's completely OK._

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,14 +1,16 @@
-# Laravel Breeze
+<p align="center"><img src="https://raw.githubusercontent.com/mnaderian/arts/main/logo-breeze.svg"></p>
 
-<a href="https://packagist.org/packages/laravel/breeze">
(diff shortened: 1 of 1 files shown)


taylorotwell:
Do you have any other ideas for the mark? Something even simpler perhaps?

m-naderian:
Let me make several designs, and send them right here.

joelbutcher:
@m-naderian How about something like this?

laravel-breeze

m-naderian:
@joelbutcher that's beautiful. But the master of the house should like it. 😉

taylorotwell:
Remember that we want absolute design consistency with the existing logos. I can have the person who usually does them do one.

mnaderian:
@taylorotwell I think this is what you wanted:

Breeze Logo

PR #16, opened 2020-11-12
[1.x] Add link Already Registered

holiq:
This adds a link so that those who have registered can login directly, like in jetstream


JeroenvanRensen:
Yeah nice idea!

PR #21, opened 2020-11-23
[1.x] Submit logout form using alpinejs

MohammadAgbareya:

### DIFF
--- a/stubs/resources/views/layouts/navigation.blade.php
+++ b/stubs/resources/views/layouts/navigation.blade.php
@@ -39,8 +39,7 @@
                             @csrf
 
                             <x-dropdown-link :href="route('logout')"
-                                    onclick="event.preventDefault();
-                                                this.closest('form').submit();">
+                                    @click.prevent="$event.target.closest('form').submit();">
                                 {{ __('Logout') }}
                             </x-dropdown-link>
                         </form>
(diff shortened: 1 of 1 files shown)


taylorotwell:
It isn't broken.

MohammadAgbareya:
> 
> 
> It isn't broken.

I tested it before pushing, it worked just fine!

MohammadAgbareya:
logout

joelbutcher:
@MohammadAgbareya He's referring to the original source code your PR replaces.

MohammadAgbareya:
> 
> 
> @MohammadAgbareya He's referring to the original source code your PR replaces. (If it ain't broke, don't fix it)

Oh, I thought he said "It is broken", I agree with you, but isn't it better to use aplinejs since you use it in the nav tag?

joelbutcher:
@MohammadAgbareya I agree, better to be consistent than not, and it's very easy to be consistent in a package this small. 

But at the end of the day, it ain't my codebase.

@taylorotwell is the one you gotta convince as to why he should replace one way of doing something that is tried, tested and working, with something else that produces the exact same effect just for the sake of 'being consistent'. I'm sure there are countless PR's like this for Laravel core, but not all of them make them in as there's just so much for he and the team to maintain and stay on top of

PR #23, opened 2020-11-26
fix confirm-password validation response

holiq:
Fix the respon on confirm-password when the password is incorrect

From this
Screenshot_20201123-012443_Chrome

To this
Screenshot_20201123-012456_Chrome


driesvints:
You need to add the new language line to your app as well: https://github.com/laravel/laravel/blob/8.x/resources/lang/en/auth.php#L17

PR #24, opened 2020-11-26
[1.x] Guards support

MohammadAgbareya:


taylorotwell:
Edit it in your own application. That's why we export it there.

MohammadAgbareya:
> 
> 
> Edit it in your own application. That's why we export it there.

I already did that in my application, I thought this might be helpful for other people who use multiple guards

PR #32, opened 2020-12-31
Allow user to specify auth guard and password broker

nathanpurcell:
For your consideration, I have taken inspiration from existing generators in Laravel to extend Breeze so that the user may pass in two arguments to specify the auth guard and password broker they wish to use. 

```bash
php artisan breeze:install 
php artisan breeze:install web users # same as above
php artisan breeze:install members members
```

In its current state, Breeze is hard-coded to support the default guard/password. While this is helpful and not too difficult to edit if required, I think it is helpful to be able to scaffold auth for a different guard as easily as the defaults. 

Naturally, the default values of the new arguments ("guard" and "passwords") are "web" and "users" respectively. 
I have therefore only set route prefixes, namespaces etc. if these arguments are not default. Much of this can be seen in the `setProperties` method. 

Unlike other Laravel generator commands, `breeze:install` was simply copying files from a stubs directory to the matching project directories. As this PR considers the user's input, we are substituting strings so I have tried to conform with past Laravel instances a renamed the stub files from `.php` files to `.stub` - I don't think this makes much difference, just preference and style. 

Obviously docs will need to be updated to reflect that the user can define their own guard/broker and that they should pass *both* arguments. 
I had considered a scenario where the user was able to pass only one argument (likely the guard) and we assumed the broker was defined with the same name but this seems unlikely and could cause confusion later. 

I have not updated tests yet as I wanted to offer this first to see if there's interest. If there is I would likely create a `MemberFactory` and local `Member` model to test against `php artisan breeze:install members members`. The tests would mostly be amended copies of the existing ones. 

The benefit to the end user is more customisable auth scaffolding in the event they wish to add an authenticated area to their application for more than just `users`. 
An additional benefit is the clearer configuration of the `Auth` and `Password` facades as both the `guard` and `broker` methods are now used even if using the default auth config. 

Happy to make amendments as directed.

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -4,6 +4,7 @@ namespace Laravel\Breeze\Console;
 
 use Illuminate\Console\Command;
 use Illuminate\Filesystem\Filesystem;
+use Illuminate\Support\Str;
 
 class InstallCommand extends Command
 {
@@ -12,7 +13,7 @@ class InstallCommand extends Command
      *
      * @var string
      */
-    protected $signature = 'breeze:install';
+    protected $signature = 'breeze:install {guard=web} {passwords=users}';
 
     /**
      * The console command description.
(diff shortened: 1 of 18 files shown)


taylorotwell:

PR #33, opened 2021-01-03
Support 2FA with google reCaptcha verification

devhoussam:
I hope laravel breeze team support 2FA with google reCaptcha verification


joelbutcher:
It looks like this PR is targeting the wrong branch?

PR #34, opened 2021-01-03
fix: update status code for request that could not be authenticated

zeravcic:
The HTTP [`401 Unauthorized`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401) client error status response code indicates that the request has not been applied because it lacks valid authentication credentials for the target resource.


taylorotwell:
Not changing this on a patch release.

PR #36, opened 2021-01-03
[1.x] Use named route for email verification screen in feature test stub

Kessir:
Email verification page url is currently hardcoded in `stubs/tests/Feature/EmailVerificationTest.php`. This PR makes use of the `verification.notice` named route.


driesvints:
This is a test, where you usually want to verify the exact matching route is called, rather than relying on a configuration.

PR #38, opened 2021-02-03
Update AuthenticatedSessionController.php

ErickZH:
Add throws tag in the docblock


driesvints:
This method does not directly throw the exception

PR #56, opened 2021-03-23
Add Route::group

Assimilationstheorie:
Add Route::group for more readability and less duplicate code.


taylorotwell:
I think it's OK as is.

Assimilationstheorie:
I'm a little confused because it's also my "first" post in the form. I use, the very useful and recommended by Laravel, systematics to group routes. This ensures that no route is called "just like that" without the protection, reduces the code and divides it sensibly. And it is simply closed, with a justification "it is good as it is".

Sorry, but that's somehow not really the basic idea of open source either. If it is already rejected, then at least a justification would be good, because I have also taken time to make this contribution...

PR #58, opened 2021-04-06
Add more robust test for registration flow

JohnnyWalkerDesign:
Add the `test_users_can_register_and_then_authenticate_using_the_login_screen()` method to ensure a user can register and then log in.

**More information**
The current suite tests to see if a user can register and if another user can login. It does not test if a user can register and _then_ login. Consider a situation where the developer has accidentally left `'password'` out of the User model `$fillable` array. All tests would pass successfully, but the app would be fundamentally broken.

This adds a more robust authentication test in situations where it's possible to collect a user's email address without a password (eg. an SCV which include mailing list subscription information). As happened to me today.

### DIFF
--- a/stubs/default/tests/Feature/AuthenticationTest.php
+++ b/stubs/default/tests/Feature/AuthenticationTest.php
@@ -5,11 +5,13 @@ namespace Tests\Feature;
 use App\Models\User;
 use App\Providers\RouteServiceProvider;
 use Illuminate\Foundation\Testing\RefreshDatabase;
+use Illuminate\Foundation\Testing\WithFaker;
 use Tests\TestCase;
 
 class AuthenticationTest extends TestCase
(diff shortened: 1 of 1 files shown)


taylorotwell:
Feel free to add to your own application if you wish.

JohnnyWalkerDesign:
> Feel free to add to your own application if you wish.

What does this mean?

driesvints:
Copy this test in your own app

JohnnyWalkerDesign:
Wow. I know you probably didn't mean it to sounds patronising but it reads like you're giving someone permission to use their own code in their own app...

JohnnyWalkerDesign:
Oh wow. Looking at this code again, I clearly wasn't paying attention. I just cut and paste without thinking. Doh.

PR #62, opened 2021-04-22
Fix Vue warning

ahmedkandel:
When visiting forgot-password page there is a console [Vue warn]: Extraneous non-props attributes (errors, auth) were passed to component but could not be automatically inherited because component renders fragment or text root nodes.
The reason is [Attribute Inheritance on Multiple Root Nodes](https://v3.vuejs.org/guide/component-attrs.html#attribute-inheritance-on-multiple-root-nodes)

Inertia shares `errors` as a prop by default and Breeze shares `auth` as a prop by default. If those props are not explicitly defined on each page component a Vue runtime warning will be issued.

We can wrap all the root nodes of the page component in a `<div>` to create a single root component and get rid of the warning, But those props will be inherited as [Non-Prop Attributes](https://v3.vuejs.org/guide/component-attrs.html#non-prop-attributes) resulting in `<div errors="[object Object]" auth="[object Object]">`

Suggested solution : explicitly define `errors` and `auth` as props on each page component.

### DIFF
--- a/stubs/inertia/resources/js/Pages/Auth/ResetPassword.vue
+++ b/stubs/inertia/resources/js/Pages/Auth/ResetPassword.vue
@@ -43,7 +43,9 @@
         },
 
         props: {
+            auth: Object,
             email: String,
+            errors: Object,
             token: String,
         },
 
--- a/stubs/inertia/resources/js/Pages/Auth/ConfirmPassword.vue
+++ b/stubs/inertia/resources/js/Pages/Auth/ConfirmPassword.vue
@@ -36,6 +36,11 @@
             BreezeValidationErrors,
         },
 
+        props: {
+            auth: Object,
+            errors: Object,
+        },
+
         data() {
             return {
                 form: this.$inertia.form({
--- a/stubs/inertia/resources/js/Pages/Auth/ForgotPassword.vue
+++ b/stubs/inertia/resources/js/Pages/Auth/ForgotPassword.vue
@@ -41,7 +41,9 @@
         },
 
         props: {
-            status: String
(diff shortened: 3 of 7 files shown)


reinink:
Yeah this solution is good, although it kind of sucks if you add additional top-level shared props.

But maybe that's the point...you probably shouldn't be constantly adding top-level shared props. Have `auth`, and `errors`, and try to stick to that, and be aware that any time you add additional top-level shared props, they will need to be explicitly set this way as well.

stubs/inertia/resources/js/Pages/Auth/ResetPassword.vue
```diff
@@ -43,7 +43,9 @@
         },
 
         props: {
+            auth: Object,
             email: String,
+            errors: Object,
```
reinink, line 48: Would it maybe be smart to always list the top level props (`auth` and `errors`) first? 🤔

ahmedkandel: I was thinking about this but as all breeze components list props in alphabetical order, so I followed the same order. Maybe when developers create their own pages they may prefer to list (`auth `and `errors`) first.

reinink: Yeah fair enough. 👍

reinink:
So, I think I figured out the right away to fix this issue. 🙌

See here: https://github.com/inertiajs/inertia/pull/681

ahmedkandel:
@reinink you are right disabling inheritance by default is much cleaner.

PR #64, opened 2021-04-29
Add jsconfig.json for better DX in VSCode.

ninjaparade:
This PR adds a `jsconfig.json` file to the Inertia scaffolding to allow javascript services in VSCode to allow better support for the `@` alias. 

image


driesvints:
Heya, can you send this to 1.x?

PR #68, opened 2021-05-11
Use boolean() instead of filled()

reinink:
This PR updates the `LoginRequest` to use the request `boolean()` method instead of `filled()` when determining if the user should be remembered.

The primary issue with `filled()` is that it will return `true` if `remember` is `false` (in a JSON request). This is common when working with Inertia where you have a boolean `remember` data value.

Using `boolean()` works for the Blade approach as well, since the underlying `FILTER_VALIDATE_BOOLEAN` check returns `true` for the value `"on"`.


driesvints:
@reinink does the boolean change still works for apps which have exported their views?

PR #80, opened 2021-05-23
[1.x] Change as from 'a' to 'button'

JoshSalway:
https://inertiajs.com/links
Links are suggested to be "button" rather than "a"

Otherwise, a warning will show by default.

@reinink FYI thoughts?

### DIFF
--- a/stubs/inertia-react/resources/js/Components/ResponsiveNavLink.js
+++ b/stubs/inertia-react/resources/js/Components/ResponsiveNavLink.js
@@ -1,7 +1,7 @@
 import { InertiaLink } from '@inertiajs/inertia-react';
 import React from 'react';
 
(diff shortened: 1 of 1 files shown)


driesvints:
Hey @JoshSalway. This seems like a breaking change to me. What happens to all apps that are expecting a link atm? Also: It's called `ResponsiveNavLink` not `ResponsiveNavButton` so I'd definitely expect a link by default.

JoshSalway:
Ok, sure. Wasn't sure why it shows a warning by default...

taylorotwell:
@driesvints it's not really possible to have a breaking change in Breeze. It is a scaffolding exporter.

taylorotwell:
Fixed by adding `as="button"` to React logout button. Thanks.

driesvints:
@taylorotwell urgh, I keep forgetting about that 🤦‍♂️

PR #81, opened 2021-05-28
remove double session regeneration

jshah4517:
On login, the session is regenerated twice. The second time appears to be unnecessary and means if you try to capture the user's session ID on the `Illuminate\Auth\Events\Login` event, it is not the right one.

The first regeneration happens in the `SessionGuard` login, it calls `UpdateSession` - https://github.com/laravel/framework/blob/6.x/src/Illuminate/Auth/SessionGuard.php#L450


taylorotwell:
Feel free to remove from your own application if you don't want it.

jshah4517:
@taylorotwell I was just wondering what the significance of the line is? The first regeneration should take care of session fixation?

PR #87, opened 2021-06-28
Add dynamic page title to layout components

imacrayon:
All of the default pages in Breeze have the same title which [isn't very accessible](https://www.w3.org/WAI/tips/writing/#provide-informative-unique-page-titles). I've updated the the Guest and App layouts to support a `title` prop so developers can set a unique page title. I think this will encourage better templating practices. The abstract Layout component I added also serves as a good place to define site-wide layout behavior.

The `title` prop accepts a `string` or `array` and automatically appends the App Name to the end of the page title:
`'Login'` becomes "Login | Laravel" and `['Child Page', 'Parent Page']` becomes "Child Page | Parent Page | Laravel".

### DIFF
--- a/stubs/default/App/View/Components/AppLayout.php
+++ b/stubs/default/App/View/Components/AppLayout.php
@@ -4,7 +4,7 @@ namespace App\View\Components;
 
 use Illuminate\View\Component;
 
-class AppLayout extends Component
+class AppLayout extends Layout
 {
     /**
      * Get the view / contents that represents the component.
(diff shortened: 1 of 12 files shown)


taylorotwell:
I suggest using a slot:

image

image

imacrayon:
Ok, I figured a slot was the wrong tool in this situation because we're not wanting to include any markup in the title. I can see how using a prop might be over-engineering a simple problem though.

PR #88, opened 2021-07-08
add name for vue components

ArielMejiaDev:
## Description

it adds the name property to every vue component that is used in breeze

## How it help the end user

it helps the IDE to autocomplete and import automatically the breeze components to use it on any other form, page component.


taylorotwell:
Feel free to add to your own application.

PR #90, opened 2021-07-20
[Inertia] Update <Link> tags & implement <Head> management (title tag)

claudiodekker:
- Adjust changes as done in https://github.com/laravel/jetstream/pull/832
- Implements <Head>-management support as to allow for per-page title tags.
- Moves forms etc. to a 'Partials' subfolder, as to more clearly indicate which files are "Pages" and which aren't.


stubs/inertia-vue/resources/js/Components/DropdownLink.vue
```diff
@@ -1,5 +1,15 @@
 <template>
-    <inertia-link class="block w-full px-4 py-2 text-left text-sm leading-5 text-gray-700 hover:bg-gray-100 focus:outline-none focus:bg-gray-100 transition duration-150 ease-in-out">
+    <Link class="block w-full px-4 py-2 text-left text-sm leading-5 text-gray-700 hover:bg-gray-100 focus:outline-none focus:bg-gray-100 transition duration-150 ease-in-out">
```
joshuapack, line 2: This is missing a href field. We will get an error like similar to this https://github.com/inertiajs/inertia/issues/744 id no href field on a Link.

pro-kurt (commented):
Identifier 'Link' has already been declared.

PR #93, opened 2021-07-26
Update app layout to make optional for header.

irsyadadl:
Give optional to the header so we can make component without header required.


taylorotwell:
Customize your own application.

PR #97, opened 2021-08-19
Make .js to .jsx for react.

irsyadadl:


driesvints:
Please add a thorough explanation why this is needed.

irsyadadl:
So that the React logo is visible when we dev, so we know which is a regular js file and a jsx file.

taylorotwell:
Feel free to change your own project.

PR #107, opened 2021-11-17
Change minimum Sanctum version to 2.8

benjivm:
Enables installation in PHP 8 projects.


driesvints:
Why doesn't the current code work for you?

PR #109, opened 2021-12-01
Breeze "API" / SPA Stack

taylorotwell:
This adds a new `api` "stack" to Breeze. This stack removes all JS and CSS related scaffolding and installs controllers that properly handle the API authentication (powered by Sanctum) that would typically be required when authenticating a JavaScript SPA such as applications powered by Next.js / Nuxt.js, etc.

This stack can be paired with an SPA of your choosing or with an upcoming Laravel Breeze implementation in Next.js 12 which will be open sourced at https://github.com/laravel/breeze-next within the coming days.

```bash
composer require laravel/breeze 

php artisan breeze:install api
```


mul14:
These files should be removed:
- `stubs/default/.DS_Store`
- `stubs/default/pest-tests/.DS_Store`

PR #111, opened 2021-12-20
Add orchestra/testbench as dev dependency

josepostiga:
This PR adds `orchestra/testbench` package as a dev dependency to allow for IDEs to have access to Laravel's class/functions declarations.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,5 +38,8 @@
         "sort-packages": true
     },
     "minimum-stability": "dev",
-    "prefer-stable": true
+    "prefer-stable": true,
+    "require-dev": {
+        "orchestra/testbench": "^6.23"
+    }
 }


driesvints:
We don't use testbench in this package so I'm not sure why this is needed?

josepostiga:
Well, when I checked out this code and installed the dependencies with Composer, I noticed PHPStorm didn't have access to some methods that were used, like the `Command`'s `option()` and `argument()` methods (for example).

Maybe I'm overkilling it by adding testbench, and probably could explicitly add the "missing" components, but that will allow for accessing (for dev purposes) to more of Laravel's components.

Does that make sense? :)

driesvints:
I think you're misunderstanding this package a bit. This package is a scaffold for a new Laravel app's auth system. Adding Testbench to this package will have no effect since we don't use Testbench in Breeze has no test suite of its own. 

You're free however, to add this package to your own app.

PR #112, opened 2021-12-23
Fix Inertia Controllers @return tag and Inertia Vue Input Component

thewebartisan7:
- Fix correct @return tag for Inertia response

- Fix Inertia vue Input component autofocus. The method `focus()` doesn't work. With this change the method is called if input has attribute "autofocus" on mounted.


taylorotwell:
I probably need to look at that autofocus thing separately.

thewebartisan7:
You can see that with React works fine when you open login, register or forgot password. But not with Vue.

PR #115, opened 2022-01-02
Use jsx instead of js

eumatv:
As we already know, react has its own file extesion jsx, and having said that, it makes a lot of sense to use it for pages and components instead of .js alone. Also, some code editors such as VS Code by default doesn't enable Emmet for js files and requires further configuration.


taylorotwell:
Most Next.js examples seem to use `.js` and it seems fairly common to use either one? Will just keep it like it is for now.

PR #117, opened 2022-01-05
Fix API Controllers @return tag

MohmmedAshraf:
This pull request fixes the DocBlocks `@return` tag for API scaffolding

- Use `JsonResponse` instated of `RedirectResponse`


taylorotwell:
Thanks for contributing to Laravel! ❤️

PR #118, opened 2022-01-09
Use FRONTEND_URL port in Sanctum config

azurinspire:
If `FRONTEND_URL` has some other port in local development than default `:3000` it is dropped out in `sanctum.php` and cookies do not work. 

Other option would be change the [documentation](https://laravel.com/docs/8.x/starter-kits#breeze-and-next) to have a notice of including the host in SANCTUM_STATEFUL_DOMAINS.

### DIFF
--- a/stubs/api/config/sanctum.php
+++ b/stubs/api/config/sanctum.php
@@ -20,7 +20,7 @@ return [
         env('FRONTEND_URL')
             ? ','.implode(':', \Illuminate\Support\Arr::whereNotNull([
                 parse_url(env('FRONTEND_URL'), PHP_URL_HOST),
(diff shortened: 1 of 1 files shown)


driesvints:
You just set it through `FRONTEND_URL`?

azurinspire:
> You just set it through `FRONTEND_URL`?

If you add `FRONTEND_URL=http://localhost:3001` to `.env` the port is not coming to `sanctum.php`, current code adds just `localhost` there and sanctum cookie does not work.

taylorotwell:
Just change this to whatever you need it to be in your own application.

azurinspire:
I don't need this anymore, I was trying to help others to not loose time with the same problem because documentation doesn't explain this at all.

PR #123, opened 2022-01-27
Taking the advantage of the new "rtl" modifier.

alqahtani:
Taking the advantage of the new "rtl" modifier to make the template support RTL when adding just dir="rtl" in layout files


driesvints:
Please see other attempts in Jetstream. Feel free to do this in your own app.

alqahtani:
> Please see other attempts in Jetstream. Feel free to do this in your own app.

@driesvints Thank you and sorry for taking your time. just a comment: the last sentence wasn't nice! I know I can use it in my own apps but I just tried to be helpful.

PR #124, opened 2022-02-03
[1.x] Use release target branch in changelog compare URL

stefanzweifel:
Use the target branch of a release in the compare URL when updating the changelog through GitHub Actions.


driesvints:
Thanks!

PR #125, opened 2022-02-07
Update app.blade.php

iosiflivadaru:
The meta tag approach for csrf tokens won't work in SPA apps, only in blade apps. Removing it fixed the issue I had with uploading files after I login!

### DIFF
--- a/stubs/inertia-common/resources/views/app.blade.php
+++ b/stubs/inertia-common/resources/views/app.blade.php
@@ -3,7 +3,6 @@
     <head>
         <meta charset="utf-8">
         <meta name="viewport" content="width=device-width, initial-scale=1">
-        <meta name="csrf-token" content="{{ csrf_token() }}">
 
         <title inertia>{{ config('app.name', 'Laravel') }}</title>
 


taylorotwell:
Hmm - how was it breaking your application?

iosiflivadaru:
I was using this config for the axios from some other apps I built. When I looked on the app.blade.php I saw that the csrf token was being set so I thought I had to do the same thing, use this config. If I don't use the config below and do a normal axios call it works just fine (noticed that after the pull request). That line of code misled me, sorry for bothering (still new to the VILT stack).
```js
const config = {
  headers: {
    "Content-Type": "multipart/form-data",
    "X-CSRF-TOKEN": document.querySelector('meta[name="csrf-token"]')
      .content,
  },
};
```

RobertBoes:
Actually, I think this line can safely be removed. Inertia / Axios uses the `X_XSRF_TOKEN` cookie by default. This is also the proper way to use the csrf token, because the cookie will be updated on every request, the meta tag won't be updated.
You'll also run into issues when trying to use the token from the meta tag, like mentioned above, since that token won't be valid

taylorotwell:
I dunno - I'm not removing this without more thorough information as to what it is breaking I don't think. I haven't had any other complaints about this breaking stuff since releasing Breeze.

RobertBoes:
> I dunno - I'm not removing this without more thorough information as to what it is breaking I don't think. I haven't had any other complaints about this breaking stuff since releasing Breeze.

It "works" because the token is not being used, problems arise when you do use the meta token, because it's an old token which isn't valid. This tag won't update between requests, so if you'd use this value you'd be sending an old token, causing an CSRF mismatch. Axios uses the cookie value (https://github.com/axios/axios/blob/73e3bdb8835ba942096b662e9441f1d85ce4d484/lib/defaults.js#L117-L118), and not the token defined in the meta tag. 

I believe the same reasoning was used to remove the tag from Jetstream https://github.com/laravel/jetstream/pull/878

driesvints:
@taylorotwell seems we indeed did the same for Jetstream?

PR #127, opened 2022-02-08
Transition Inertia Vue stubs to `<script setup>` syntax

jessarcher:
This PR transitions the Inertia Vue 3 stubs to use Vue's `<script setup>` syntax.

The [official Vue docs](https://vuejs.org/api/sfc-script-setup.html) recommend using the `<script setup>` syntax when using the Composition API with Single File Components as it reduces boilerplate and increases performance. It also significantly improves working with TypeScript, for users who may wish to convert their Breeze installation.

The PR also moves the `<script setup>` section to the top of the single file components to be consistent with the Vue docs and Evan's recommendations.

Note that the semicolons in the current code were applied inconsistently so I have also added them where appropriate. The Vue docs don't use semicolons, however the StyleCI [JS](https://docs.styleci.io/multi-language#javascript-config) and [Vue](https://docs.styleci.io/multi-language#vue-config) rules require them by default.

I have tested this PR in a new Laravel Breeze installation by viewing every page component and following every action (register, login, verify email, forgot password, password reset, logout) happy and sad path. I would love for someone else to also double check this.

### DIFF
--- a/stubs/inertia-vue/resources/js/Components/Input.vue
+++ b/stubs/inertia-vue/resources/js/Components/Input.vue
@@ -8,8 +8,8 @@ defineEmits(['update:modelValue']);
 const input = ref(null);
 
 onMounted(() => {
-    if (input.hasAttribute('autofocus')) {
-        input.focus();
+    if (input.value.hasAttribute('autofocus')) {
+        input.value.focus();
     }
 });
 </script>
--- a/stubs/inertia-vue/resources/js/Components/Input.vue
+++ b/stubs/inertia-vue/resources/js/Components/Input.vue
@@ -1,17 +1,19 @@
+<script setup>
+import { ref } from 'vue';
+
+defineProps(['modelValue']);
+
+defineEmits(['update:modelValue']);
+
+const input = ref(null);
+
+const focus = () => input.focus();
+
+defineExpose({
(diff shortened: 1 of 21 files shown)


stubs/inertia-vue/resources/js/Components/Input.vue
```diff
@@ -1,17 +1,19 @@
...
 
-    methods: {
-        focus() {
-            this.$refs.input.focus()
-        }
-    }
-}
+defineExpose({
+    focus,
+});
```
jessarcher, line 14: I'm not sure what the `focus()` method is actually there for as I can't see it being called from anywhere. I've left it in (with the new syntax) in case it's intended for users, but I'm also happy to remove it.

jessarcher: Just for documentation purposes, the focus code seems to be leftover from an attempt to make `<Input autofocus>` work. I've fixed this in d868309 and removed the `defineExpose()` as it wasn't needed.

taylorotwell:
Thanks!

mehranny:
This causing an error for me: Uncaught ReferenceError: defineProps is not defined
    at eval (Button.vue?3794:2:1)

PR #131, opened 2022-02-13
[1.x] Group common middleware instead of duplicating

dammy001:

### DIFF
--- a/stubs/default/routes/auth.php
+++ b/stubs/default/routes/auth.php
@@ -14,12 +14,12 @@ Route::middleware('guest')->group(function () {
     Route::get('register', [RegisteredUserController::class, 'create'])
                 ->name('register');
 
-    Route::post('register', [RegisteredUserController::class, 'store'])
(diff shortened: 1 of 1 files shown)


driesvints (approved):
I also like this approach 👍

stubs/default/routes/auth.php
```diff
@@ -10,55 +10,47 @@
...
-Route::post('/reset-password', [NewPasswordController::class, 'store'])
-                ->middleware('guest')
+    Route::post('reset-password', [NewPasswordController::class, 'store'])
                 ->name('password.update');
+});
 
-Route::get('/verify-email', [EmailVerificationPromptController::class, '__invoke'])
-                ->middleware('auth')
+Route::middleware('auth')->group(function () {
+    Route::get('verify-email', [EmailVerificationPromptController::class, '__invoke'])
```
2bj, line 38: maybe: `Route::get('verify-email', EmailVerificationPromptController::class)`?

PR #138, opened 2022-02-18
Append Breeze Routes Into web.php after running install command

ShakilAhmmed:


ShakilAhmmed:
> This shouldn't be necessary btw. Why do you need this?

Middle of Project, If i install breeze it replace web.php.

driesvints:
You only need to install Breeze when starting a new project. If you want to use it on an existing project you'll have to install everything manually.

PR #146, opened 2022-03-25
[1.x] Add Inertia SSR Support 

xiCO2k:
Same exact approach from https://github.com/laravel/jetstream/pull/1012

Beside Vue, we also got React working with SSR.

PS: Also updated the node dependencies.


xiCO2k:
Will add React Inertia SSR as well.

xiCO2k:
Should be good to go!

taylorotwell:
Thanks!

PR #147, opened 2022-03-25
[1.x] Grouping of the routes within their controllers

DerianCordobaPerez:
* Easy readability when reading inside paths.
* Easy scalability on the actions taken by each controller.
* Split extra properties for a certain group of routes without needing to perform more group functions.


taylorotwell:
I think it's fine as is. Thanks.

PR #149, opened 2022-04-05
update inertia stub

Loafer19:
- removed useless lines

unnecessary connection of scripts


taylorotwell:
No explanation of why this is useless.

Loafer19:
ok, maybe then someone else will explain why there is a connection to this script, why there is a hardcode localhost

ArielMejiaDev:
@Loafer19 it is useful as it prevents some issues when mix runs things like browserSync feature.

PR #152, opened 2022-04-29
Lovecoding git issue fix

lovecoding-git:
I installed the fresh Laravel V9 and Breeze react.
When I login and click dropdown, I can't click Log Out since it has `z-40` and when I remove that class, it works.

I think we can either remove `z-40` class or add some `relative z-50` to dropdown transform.

image


taylorotwell:
Hmm, this class is present on all the Breeze stubs as well without issue. Maybe just remove it from your application?

fouteox:
same here

PR #156, opened 2022-06-11
bump TailwindCSS to 3.1

geisi:
Hey Laravel Team,

this PR updates TailwindCSS to the new 3.1 release and adds the new FirstParty TypeScript types to the tailwind.config.js.

https://tailwindcss.com/blog/tailwindcss-v3-1


jessarcher:
Nice :+1:

martinvadel:
Hi, pardon me if i'm wrong but should not be the type annotation right above module.export and not above const defaultTheme? Not sure how it works in other editors but in my VSCode it works only when is it right above module.export.

geisi:
@martinvadel 
You are right thank you. I will correct this with a new PR.

PR #159, opened 2022-06-20
Allow support for non-jsx React files in Tailwind config

morris14:
Not all developers will use .jsx as the file extension for their React components. I was editing a React component recently with a .js prefix and realised my tailwind classes weren't getting generated correctly. This should help support that.


morris14:
Support for this indicated here: https://tailwindcss.com/docs/configuration

jessarcher:
Hi @morris14,

With the upcoming switch to Vite in #158, new projects will need to use a `.jsx` extension for files containing JSX as this is [required by Vite](https://twitter.com/youyuxi/status/1362050255009816577).

Do you know of any common scenarios where a `.js` file would contain Tailwind classes and _not_ be using JSX?

PR #162, opened 2022-06-25
Add Sanctum Guards section

Nir-An:
Add the original [Sanctum ability](https://github.com/laravel/sanctum/blob/2.x/config/sanctum.php#L24-L37) to change the authentication guards that will be checked when trying to authenticate a request


taylorotwell:
If you need to configure this add it to your own project.

PR #169, opened 2022-07-11
Fix: React SSR installation error

renomureza:
When I installed Breeze (React & SSR) as per this tutorial: https://laravel.com/docs/9.x/starter-kits#laravel-breeze-installation

When running the command `php artisan breeze:install react --ssr`, it returns an error like the following image.

This is because when we unlink the `app.js` file doesn't exist, I added checking for the existence of the file first.

Screenshot 2022-07-11 144626


parth391:
I am still getting bellow error

Screenshot 2022-11-08 at 7 58 56 PM

PR #171, opened 2022-07-14
[1.x] Vite 3 support

timacdonald:
## Dependent PRs

- https://github.com/laravel/laravel/pull/5944
- https://github.com/laravel/vite-plugin/pull/77

Although it may be possible to allow `'@inertiajs/server'` to be externalised, to get this across the line now I think we just move forward and can iterate if we can solve it upstream with Inertia.


jessarcher:
@timacdonald As with Jetstream, I've created a branch that keeps a single config file: https://github.com/laravel/breeze/compare/vite-3-single-config

Keen to hear your thoughts!

jessarcher (approved):

PR #175, opened 2022-08-02
[1.x] Display validation errors alongside their field

jessarcher:
This PR updates Breeze to display validation errors beneath their respective fields, instead of all grouped together at the top of the page.

The Vue stubs already contained the `InputError` component for this, but it was not being used. The React stubs were missing the component so I have added it in.

**Before**
image

**After**
image


driesvints (approved):

siarheipashkevich:
Hi @jessarcher 

Laravel returns the array of the errors per one field. Is it handled?

Thanks

jessarcher:
> Laravel returns the array of the errors per one field. Is it handled?

Inertia only returns the first error for each field.

See: https://github.com/inertiajs/inertia-laravel/blob/540b953ec383364264f9bd633849db16560a4461/src/Middleware.php#L159

PR #176, opened 2022-08-06
[1.x] Use `__invoke` for single action controllers and group common middleware

dammy001:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /vendor
+.idea
 composer.lock
 /phpunit.xml
 .phpunit.result.cache
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 /vendor
-.idea
 composer.lock
 /phpunit.xml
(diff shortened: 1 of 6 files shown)


timacdonald:
If we do this we probably want to rename the controllers to show what the action is that the take.

e.g. `NewPasswordController` > `StoreNewPasswordController` etc.

I'd prefer we only utilised single action controllers for controllers that land outside of the REST actions and fallback to a more RPC style approach. Just my thoughts.

.gitignore
```diff
@@ -1,4 +1,5 @@
 /vendor
+.idea
```
jessarcher, line 2: This should really be part of your own global `gitignore` as it's not a file generated by this repo.

https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/

dammy001: Thanks

taylorotwell:
I'm fine with how it is for now.

PR #183, opened 2022-08-29
update registration tests

ousid:
Similar to [Jetstream Registration Tests](https://github.com/laravel/jetstream/blob/2.x/stubs/tests/RegistrationTest.php#L17), This PR allows users to skip registration tests without throwing an error, if the user tried to disable the registration routes

In fact, the users can just remove the registration test file directly, but sometimes in some applications, they just disable the registration routes temporarily, in hope to get them back in the future. So, it's easy to skip tests while the registration is disabled instead of remove them entirely.


taylorotwell:
Just delete those tests if you don't need them.

PR #184, opened 2022-09-05
[1.x] Test stubs in CI

jessarcher:
This PR adds a `tests` workflow to ensure that the Breeze stacks compile successfully and that the stubbed tests pass.

The workflow is based on the [Jetstream tests workflow](https://github.com/laravel/jetstream/blob/2.x/.github/workflows/tests.yml). The main difference is that I've removed the `test` job because this repository doesn't have any tests (excluding the tests that Breeze scaffolds, which are now tested).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,58 @@
+name: tests
+
+on:
+  push:
+  pull_request:
+  schedule:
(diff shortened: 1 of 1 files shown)


driesvints (approved):

driesvints:
Nj!

PR #192, opened 2022-09-25
Upgrade to React 18

kjoedion:
The current breeze react stack uses old React 17 version. Inertia has recently released full support for React 18.

This PR simply updates the package json to react 18 and uses the react 18 `createRoot` api which is all that is required to make it work.


jessarcher (approved):
I don't know a whole lot about React, but I've tested this and it seems to work well. I've confirmed the SSR version still works too.

Thanks! :raised_hands:

PR #196, opened 2022-10-24
[1.x] Vite preloading

timacdonald:
Depends on: https://github.com/laravel/framework/pull/44096

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "illuminate/validation": "^9.21"
     },
     "conflict": {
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -21,7 +21,7 @@
         "illuminate/validation": "^9.21"
     },
     "conflict": {
-        "laravel/framework": "<9.19.0"
+        "laravel/framework": "<9.37.0"
```
timacdonald, line 24: Taking a guess that this is the version we release this in. Will need to confirm on release day.

PR #197, opened 2022-10-24
[1.x] Bump React and Vue Dependencies

dammy001:


driesvints:
These dependencies are already installed with the current constraints.

dammy001:
@driesvints The dependencies version should remain the same even when there's minor update?

PR #200, opened 2022-10-28
Add picocolors to devDepedencies

holiq:
When i install breeze with yarn, when i run yarn build i got error can't find module picocolors, after I re-installed picocolors, I can only run yarn build as shown below 
 
Screenshot_20221028-183927_Termux


driesvints:
We believe this is a bug with Vite itself. @jessarcher and @timacdonald are investigating this.

holiq:
Okay, then what should I do about this PR?

driesvints:
Wait until they respond.

timacdonald:
This is an upstream bug that has should have been addressed in Vite 3.2.1

holiq:
Okay, thanks for the response @timacdonald

PR #205, opened 2022-11-02
Add "Profile" page

jessarcher:
This PR introduces a new "Profile" page where users can update their name, email, and password. It also allows them to delete their account.

image

This is a simpler version of the profile functionality provided by Jetstream. It doesn't have as many features, but it makes Breeze more complete as users can now update the information they provided on registration. Unlike Jetstream, *all* of the functionality is stubbed out for the user without Fortify, and with no requirement to leave the Breeze package installed afterwards. It uses a typical Laravel controller and form request, making it a *breeze* to add additional fields.

This PR introduces a few new components to achieve the above, which will also be useful for users as they build out their application:

- `SecondaryButton`
- `DangerButton`
- `Modal`

I went for a simpler design than Jetstream to avoid additional components like "FormSection" and to make it easier for the user to customise, which is especially helpful when implementing a custom design.

The functionality has been introduced for the Blade, Vue, and React stacks, but not the API stack. Tests have been provided for the user as well.


.github/workflows/tests.yml
```diff
@@ -13,9 +13,15 @@ jobs:
...
       fail-fast: true
       matrix:
-        stack: [blade, react, react --ssr, vue, vue --ssr, api]
+        stack: [blade, react, vue, api]
+        args: [null, --pest]
+        include:
+          - stack: vue
+            args: --ssr
+          - stack: react
+            args: --ssr
```
jessarcher, line 22: This updates CI to test the Pest version for each stack.

jessarcher, line 22: This updates CI to test the Pest version for each stack.

.github/workflows/tests.yml
```diff
@@ -13,9 +13,15 @@ jobs:
...
 
-    name: Test Stubs (${{ matrix.stack }})
+    name: Test Stubs - ${{ matrix.stack }} ${{ matrix.args }}
```
jessarcher, line 24: I changed the formatting here just so that when `matrix.args` is null, the whitespace before it is not obvious.

jessarcher:
Much of the front-end component code came from Jetstream so it's tried and tested. The React components needed to be made from scratch as Jetstream doesn't have React, and the Blade modal required some tweaks to work without Livewire (Jetstream comes with Livewire, Breeze is plain Blade).

taylorotwell:
@jessarcher do you think a "Saved." (similar to Jetstream) indicator would be beneficial? I notice there is no status notification if I successfully update my password for example. The page just reloads.

jessarcher:
> @jessarcher do you think a "Saved." (similar to Jetstream) indicator would be beneficial? I notice there is no status notification if I successfully update my password for example. The page just reloads.

Definitely! The Vue and React versions have it already using Inertia's `form.wasRecentlySuccessful`. I wanted it for the Blade version as well, but it's a little tricky to display on the correct form as they both (currently) use the same endpoint.

Some ideas:

1. Combine the two forms into one, making sure the validation and controller allow for the form to be submitted without supplying any password details to leave them unchanged.
2. Separate the endpoints for each form so we can flash a unique key/value.
3. Add a hidden field to each form that we can use to flash a unique key/value.
4. Livewire?

taylorotwell:
My hunch would be to have a separate controller for updating profile information and a separate controller for updating the password.

jrd-lewis:
If the profile page is mostly being brought from Fortify / Jetstream, should it also be brought over to the laravel/ui package for feature parity?

driesvints:
@jrd-lewis we have no plans to further develop laravel/ui, only maintenance updates.

jessarcher:
> My hunch would be to have a separate controller for updating profile information and a separate controller for updating the password.

Done :grinning:

stubs/default/routes/auth.php
```diff
@@ -31,7 +32,7 @@
                 ->name('password.reset');
 
     Route::post('reset-password', [NewPasswordController::class, 'store'])
-                ->name('password.update');
+                ->name('password.store');
```
jessarcher, line 35: This was renamed because the `update` method on the new `PasswordController` is more deserving of this name.

PR #207, opened 2022-11-14
Update web.php

irsyadadl:
Implement route controller for `ProfileController`.

### DIFF
--- a/stubs/default/routes/web.php
+++ b/stubs/default/routes/web.php
@@ -22,10 +22,10 @@ Route::get('/dashboard', function () {
     return view('dashboard');
 })->middleware(['auth', 'verified'])->name('dashboard');
 
-Route::middleware('auth')->group(function () {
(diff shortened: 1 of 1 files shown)


stubs/default/routes/web.php
```diff
@@ -22,10 +22,10 @@
     return view('dashboard');
 })->middleware(['auth', 'verified'])->name('dashboard');
 
-Route::middleware('auth')->group(function () {
-    Route::get('/profile', [ProfileController::class, 'edit'])->name('profile.edit');
-    Route::patch('/profile', [ProfileController::class, 'update'])->name('profile.update');
-    Route::delete('/profile', [ProfileController::class, 'destroy'])->name('profile.destroy');
+Route::controller(ProfileController::class)->middleware(['auth', 'verified'])->group(function () {
```
driesvints, line 25: Adding the verified method here, is not needed.

irsyadadl: I should not using verified here.

PR #214, opened 2022-11-16
React formatting

fouteox:
Cleaning up unused code


stubs/inertia-react/resources/js/Pages/Profile/Partials/UpdateProfileInformationForm.jsx
```diff
@@ -35,7 +35,6 @@ export default function UpdateProfileInformation({ mustVerifyEmail, status, clas
 
                     <TextInput
                         id="name"
-                        type="text"
```
jessarcher, line 38: I know this is the default, but it's included in other places like Register and Login.

fouteox: Yes I was not sure to delete that.

PR #216, opened 2022-11-21
Add Confirm Password in API

joaopalopes24:
This PR adds the route to confirm the password if the user wants to use it in his API.

It's a feature that Laravel Breeze already has, except in its API option.

Tests have been added.


joaopalopes24:
@taylorotwell This functionality already comes by default in Laravel Breeze, both in the Blade, Vue and React option. Why not add it to the API option?

joaopalopes24:
I believe it will be a good contribution.

PR #217, opened 2022-11-21
Add Confirm Password in API

joaopalopes24:
This PR adds the route to confirm the password if the user wants to use it in his API.

It's a feature that Laravel Breeze already has, except in its API option.

Tests have been added.

@taylorotwell This functionality already comes by default in Laravel Breeze, both in the Blade, Vue and React option. Why not add it to the API option?

I believe it will be a good contribution.

### DIFF
--- a/stubs/api/app/Http/Controllers/Auth/ConfirmablePasswordController.php
+++ b/stubs/api/app/Http/Controllers/Auth/ConfirmablePasswordController.php
@@ -3,7 +3,6 @@
 namespace App\Http\Controllers\Auth;
 
 use App\Http\Controllers\Controller;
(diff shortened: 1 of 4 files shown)


driesvints:
Hi @joaopalopes24, like @taylorotwell said in https://github.com/laravel/breeze/pull/216 we're not planning on this ourselves right now. Feel free to just add this to your app.

joaopalopes24:
@driesvints Taking advantage of the discussion, I wanted to ask if the breeze team has the objective of adding 2FA in the package. I think it would be an interesting feature.

driesvints:
Thank you but right now we have no plans on extending Breeze further.

PR #218, opened 2022-11-24
Fix close modal with escape keydown and cleanup import react

fouteox:
Hi, I fixed modal closing by adding "show" in useEffect dependency on DeleteUserForm page.

Then I cleaned up React calls that are no longer needed.

### DIFF
--- a/stubs/inertia-react/resources/js/Components/TextInput.jsx
+++ b/stubs/inertia-react/resources/js/Components/TextInput.jsx
@@ -1,7 +1,7 @@
 import { forwardRef, useEffect, useRef } from 'react';
 
 export default forwardRef(function TextInput(
-    { type = 'text', name, value, className, autoComplete, required, isFocused, handleChange },
+    { type = 'text', name, id, value, className, autoComplete, required, isFocused, handleChange },
     ref
 ) {
     const input = ref ? ref : useRef();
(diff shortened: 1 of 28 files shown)


fouteox:
add fix for label input

stubs/inertia-react/resources/js/Components/TextInput.jsx
```diff
@@ -17,6 +17,7 @@ export default forwardRef(function TextInput(
             <input
                 type={type}
                 name={name}
+                id={name}
```
jessarcher, line 20: I'm not a fan of re-using the `name` prop here as sometimes they need to be different.

I think it should be `id={id}` with a new `id` prop added to the component definition and everywhere the component is used with a label.

jessarcher (changes requested):
Hey @fouteox,

Thanks for the PR. I've confirmed this fixes the close on escape and label behaviour, although I've left a comment regarding the latter.

As for the `React` import, I'm not super familiar with React and I appreciate any improvements that can be made.

I've tried your PR though, and I get the following warning during the SSR build: 

image

You can see it in CI [here](https://github.com/laravel/breeze/actions/runs/3542345661/jobs/5947734957#step:7:41) as well.

And when I log in to the application, nothing renders on the dashboard route and I get this error in the browser console:

image

I would like to get the other changes merged so it might be better to move the import changes to a separate PR if you'd still like to pursue it.

Thanks again!

stubs/inertia-react/resources/js/Components/TextInput.jsx
fouteox: I have add id in all component used TextInput.jsx

fouteox:
Strange, I tried to compile on my side and I don't have this error, I sent a new commit by deleting the last import remaining in ssr.jsx, can you try again please?

fouteox:
I finally found the problem with the React import, in the Dropdown file, so I fixed it.

I also refactored the Modal component using the power of headlessui which natively creates a portal and supports the keyboard.

jessarcher (approved):
Thanks, @fouteox! This looks and works great!

You've improved the functionality _and_ reduced the code :tada:

PR #219, opened 2022-11-25
status must be String

marsuboss:
The variable status must be string because in component UpdateProfileInformationForm, status is string.


jessarcher (approved):
Thanks!

driesvints:
Thanks!

PR #220, opened 2022-11-25
Update NewPasswordController.php redirect to new helper function

andyunleashed:
Update the `redirect()->route` to use the new Laravel 9 helper function `to_route` instead. Came across it when I did a search on one of my apps to update old references to `redirect()->route`


driesvints:
Feel free to do this in your own app 👍

PR #223, opened 2022-11-28
Double quotes fix in the translation string

TENIOS:


TENIOS:
Sorry, i forgot to add backslash...

PR #228, opened 2022-12-09
[1.x] Blade stack - Moved duplicated logo component to guest layout

magdicom:
Moved duplicated logo component to guest layout


taylorotwell:
To bring consistency with Vue / React stacks I don't think we need `auth-card` either. I have removed it.

PR #232, opened 2022-12-16
Renamed isFocused prop on TextInput component

drewmw5:
autofocus does not exist as a prop on the [TextInput](https://github.com/laravel/breeze/blob/1.x/stubs/inertia-react/resources/js/Components/TextInput.jsx) component. Renaming the prop to isFocused results in the text input auto focusing. 

Thanks.


timacdonald (approved):
Confirmed issue exists, only impacts this page, and is not an issue for the Vue stack.

This fixes it. Thanks!

driesvints:
Thanks!

PR #233, opened 2022-12-19
add typescript support for intertia vue stack

moh-slimani:
I worked on few projects where i wanted to use typescript and this is how i configured it to work on the Inertia Vue stack
and i wanted to share it with other developers because it can take a lot of time to convert everything to typescript each time anyone work on a new project 

this doesn't break any existing features because i have add new folder for the stubs and new functions in the  PHP Treats to make this happen 

I added the typescript support only to vuejs because i'm not familiar with react

> there are few statements where i have added `@ts-ignore` to them , i couldn't convert those statements sorry about that but everything should works fine


taylorotwell:
Hi! I suggest forking this into your own repository like "your-username/breeze-ts"

PR #236, opened 2023-01-09
[1.x] Prompt when the stack argument is not provided

jessarcher:
This PR updates the installer to not have a default stack.

Instead, when the `stack` argument is not provided and the terminal is interactive, the installer will prompt the user:

image

It will then ask whether they would like to install dark mode, SSR (if applicable), and/or Pest tests. These all default to no.

image

> **Note** If the user provides the `stack` argument, then it's assumed they know what they want, and the command won't prompt for anything.

When run in non-interactive mode, and no `stack` argument is provided, the command will error:

image

This change could technically be considered a breaking change, but given Breeze is a single-use package, I don't think it's worth a major version bump.

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -6,6 +6,8 @@ use Illuminate\Console\Command;
 use Illuminate\Filesystem\Filesystem;
 use Illuminate\Support\Str;
 use RuntimeException;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Finder\Finder;
 use Symfony\Component\Process\PhpExecutableFinder;
 use Symfony\Component\Process\Process;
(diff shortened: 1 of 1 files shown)


src/Console/InstallCommand.php
```diff
@@ -55,6 +64,34 @@ public function handle()
...
+     *
+     * @param  \Symfony\Component\Console\Input\InputInterface  $input
+     * @param  \Symfony\Component\Console\Output\OutputInterface  $output
+     * @return void
+     */
+    protected function interact(InputInterface $input, OutputInterface $output)
+    {
+        if ($this->argument('stack') === null && $this->option('inertia')) {
+            $input->setArgument('stack', 'vue');
+        }
```
jessarcher, line 78: Do you think we can just remove the deprecated `--inertia` option instead of this?

driesvints: Think that will give errors for people who are using it in automated workflows. Might be best to wait for a major release to do that.

PR #242, opened 2023-01-29
Add macro to InstallCommand

guizoxxv:
Following the idea behing `laravel/ui` package [presets](https://laravel.com/docs/contributions#which-branch), I believe that it would be benefitial to allow users from the community to create macros to extend the funcionality of the `laravel/breeze` package.

I understand that it is not viable to include support to a huge number of stacks in the package, but with this, users outside the Laravel development team could create and maintain presets. For example, I created this [guizoxxv/laravel-breeze-bootstrap](https://github.com/guizoxxv/laravel-breeze-bootstrap) package to support Bootstrap.

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -49,6 +49,10 @@ class InstallCommand extends Command
      */
     public function handle()
     {
+        if (static::hasMacro($this->argument('stack'))) {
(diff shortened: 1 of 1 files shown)


guizoxxv:
Got it. Thanks for the feedback.

PR #244, opened 2023-02-01
Automatically send an email verification notification when a user changes their email

gpluess:
As a user I'd expect to automatically receive a verification email after I'm told to verify my email address.

This PR adds a call to the sendEmailVerificationNotification() method after a user's email has been set to null.


gpluess:
Nevermind :) I didn't look close enough to the example implementation

catalin-bratu:
@gpluess, could you explain your Eurika moment? I'm also thinking that it should automatically send it.

gpluess:
@catalin-bratu I think I just added that line to my ProfileController. The stubs get published after you install Breeze, so basically you can add any functionality you need.

PR #245, opened 2023-02-02
fix: use autocomplete for all fields for password manager compatibility

lukearmstrong:
Testing a Laravel application earlier I noticed the password manager in Chrome wasn't able to set a new password on the registration form, it only completed one of the fields with the generated password.

image

But it was able to on the edit profile form, so then I just double checked all of the inputs to make sure they all had the correct autocomplete values to hint to browsers and password managers what to do.

### DIFF
--- a/stubs/default/resources/views/auth/login.blade.php
+++ b/stubs/default/resources/views/auth/login.blade.php
@@ -8,7 +8,7 @@
         <!-- Email Address -->
         <div>
             <x-input-label for="email" :value="__('Email')" />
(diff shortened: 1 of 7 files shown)


driesvints:
This looks neat. Have you tried this with other use cases as well? (Apple Keychain, 1Password, iOS)?

taylorotwell:
@driesvints I think this is pretty standard 👍

driesvints:
Cool, @lukearmstrong can you also send this to Jetstream?

lukearmstrong:
> Cool, @lukearmstrong can you also send this to Jetstream?

Yep, will check the inputs there too.

PR #246, opened 2023-02-06
[1.x] Make CJS files explicit

timacdonald:


stubs/inertia-common/tailwind.config.js
```diff
@@ -1,7 +1,8 @@
-const defaultTheme = require('tailwindcss/defaultTheme');
+import defaultTheme from 'tailwindcss/defaultTheme';
+import forms from '@tailwindcss/forms'
```
timacdonald, line 2: 📌 `'tailwindcss/defaultTheme'` does not require, or work with, a prefixed `@` here

PR #256, opened 2023-02-13
Don't redirect in api mode if authenticated

renatokira:
A proposed improvement for the laravel/breeze package is for when it is installed in API mode, there will not be a redirect if the user is already authenticated. This change will ensure a better user experience.

It's possible to implement my own solution in Laravel, but if Breeze is installed with the --api flag, I think it would be appropriate to come this way.

## Screenshots
image
image


taylorotwell:
Hey there - you can add this to your own application if you wish.

PR #257, opened 2023-02-15
[1.x] Improve React components

jessarcher:
This PR makes the following changes to the React components (and their usage throughout the application):

* Additional props are now passed to the underlying HTML element (similar to Vue). This allows us to remove explicit props for things like `id`, `autoComplete`, and `required`.
* The `forInput` prop has been removed in favour of passing the standard JSX `htmlFor` through.
* The `handleChange` prop on `TextInput` has been removed in favour of passing the standard JSX `onChange` through.
* The `processing` prop on buttons has been changed to `disabled` to reflect better what it's doing and also to match the Vue stack.
* All components customising the `className` of an underlying element now accept a `className` prop that will be appended to the underlying element. Previously only some components were doing this.
* Explicit props that specify a redundant default (i.e. the same default as the underlying component) have been removed.


stubs/inertia-react/resources/js/Components/Dropdown.jsx
```diff
@@ -70,13 +70,11 @@ const Content = ({ align = 'right', width = '48', contentClasses = 'py-1 bg-whit
...
-const DropdownLink = ({ href, method, as, children }) => {
+const DropdownLink = ({ className, children, ...props }) => {
     return (
         <Link
-            href={href}
-            method={method}
-            as={as}
-            className="block w-full px-4 py-2 text-left text-sm leading-5 text-gray-700 dark:text-gray-300 hover:bg-gray-100 dark:hover:bg-gray-800 focus:outline-none focus:bg-gray-100 dark:focus:bg-gray-800 transition duration-150 ease-in-out"
+            {...props}
+            className={`block w-full px-4 py-2 text-left text-sm leading-5 text-gray-700 dark:text-gray-300 hover:bg-gray-100 dark:hover:bg-gray-800 focus:outline-none focus:bg-gray-100 dark:focus:bg-gray-800 transition duration-150 ease-in-out ${className}`}
```
lisatassone, line 77: `className` can be undefined and will print that in this expression, so might want to handle that case.

PR #258, opened 2023-02-15
Remove pages ref from Vite in Vue

paulocastellano:
The inertia docs https://inertiajs.com/server-side-setup, recommend uses ```@Vite('resources/js/app.js')```.

The old way creates 502 errors because cannot render complex components on the build.

### DIFF
--- a/stubs/inertia-vue/resources/views/app.blade.php
+++ b/stubs/inertia-vue/resources/views/app.blade.php
@@ -12,7 +12,7 @@
 
         <!-- Scripts -->
         @routes
-        @vite(['resources/js/app.js', "resources/js/Pages/{$page['component']}.vue"])
+        @vite('resources/js/app.js')
         @inertiaHead
     </head>
     <body class="font-sans antialiased">


taylorotwell:
@jessarcher @timacdonald ping

timacdonald:
Removing this could create a waterfall of requests for applications. I would personally prefer we didn't make this change the default.

Could you elaborate on, and potentially provide a reproduction repository, for the issue you are facing?

The following works as expected with no `502` errors:

```
composer create-project laravel/laravel breeze-test \
  && composer require laravel/breeze \
  && php artisan breeze:install vue \
  && php artisan serve
```

jessarcher:
You'd only remove this if switching to eager loading instead of code-splitting.

Inertia recommends eager loading, which creates one big bundle that is downloaded on the first request with no additional downloads on subsequent requests. Any changes to your code will create a fresh bundle that must be re-downloaded.

In Breeze, we're using code-splitting, which creates separate modules so that the first page load is faster and additional modules are only downloaded as needed. Any changes to your code will only affect the modules that are involved, improving the caching. Additional modules can also be preloaded, so any additional downloads will happen in the background instead of when a page is loading.

Each method has advantages and drawbacks, but code-splitting is the default in Vite and is our current preference.

paulocastellano:
> Removing this could create a waterfall of requests for applications. I would personally prefer we didn't make this change the default.
> 
> Could you elaborate on, and potentially provide a reproduction repository, for the issue you are facing?
> 
> The following works as expected with no `502` errors:
> 
> ```
> composer create-project laravel/laravel breeze-test \
>   && composer require laravel/breeze \
>   && php artisan breeze:install vue \
>   && php artisan serve
> ```

I will create the production environment for you to check the bug today, I ping you when it's ready.

More info it's: the bug will happens only if you using a load balancer.

PR #259, opened 2023-02-15
[1.x] Laravel 10.x adjust wording

TENIOS:


TENIOS:
This also needs to be done for the rest of the stacks.

PR #261, opened 2023-02-17
Added laravel breeze typescript scaffolding option

JensvandeWiel:
This pull request adds the --ts option to laravel breeze, this way people can generate a laravel project and breeze with typescript from the beginning without needing to do it manually.


taylorotwell:
Hey there - we are actually already working on this internally. I'm going to share this with the team to compare implementations.

jessarcher:
Thanks for your efforts on this! My concern with creating dedicated stubs for the TS versions is that it becomes a bigger maintenance burden and much more work to add new Breeze features.

I'm currently exploring an approach where we only need to maintain TypeScript stubs and the installer will just strip out the type information when TypeScript support hasn't been selected.

JensvandeWiel:
The only thing for react is to change to tsx and add the right tsconfig file for vue you only have to change the script language to ts

PR #277, opened 2023-03-21
Fix ref in react TextInput

denis-n-ko:
Fix ref in react TextInput


jessarcher (approved):
Thanks!

It looks like I broke this in #270.

driesvints:
Thanks!

PR #279, opened 2023-03-28
[1.x] Fix Pest test installation for Inertia stack

jessarcher:
This PR fixes an issue in the Breezes installer where it would always install the PHPUnit version of `ProfileTest` and `PasswordUpdateTest` regardless of whether the `--pest` option was specified.

Thanks go to [this tweet](https://twitter.com/swazlo/status/1639623260852850689) for raising the issue.

I've also added a missing failure handler to the React stack installer for when Pest fails to be installed, which I noticed was present in the other stack installers.

### DIFF
--- a/src/Console/InstallsInertiaStacks.php
+++ b/src/Console/InstallsInertiaStacks.php
@@ -88,7 +88,11 @@ trait InstallsInertiaStacks
             return 1;
         }
 
-        (new Filesystem)->copyDirectory(__DIR__.'/../../stubs/inertia-common/tests/Feature', base_path('tests/Feature'));
+        if ($this->option('pest')) {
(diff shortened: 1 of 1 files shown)


LTroya:
Hi @jessarcher, maybe this is not the best place to ask this. 

But have you tried to run the test suite created by Breeze on new projects on Github Actions? Some of them pass, while others fail.

I asked the question on [Laracast](https://laracasts.com/discuss/channels/laravel/failing-tests-when-running-pest-in-github-actions) too. 

Maybe I am doing something wrong but I haven't found the issue.

Cheers.

jessarcher:
Hi @LTroya,

We have GitHub Actions on this repository, which installs each of the Breeze stacks in a fresh Laravel application and runs the test suite ([example](https://github.com/laravel/breeze/actions/runs/4538487365/jobs/7997438318?pr=279)). I've never seen any issues with some of the tests failing as they do in your example.

I've left an additional reply on your Laracasts post that I hope helps.

PR #280, opened 2023-04-09
Remove redundant form data in vue stub

datlechin:
Remove redundant form data in vue stub


ArielMejiaDev:
Hi @taylorotwell, I think that it would be better to keep it as the `remember` key, which sets explicitly a value that could be changed, it adds visibility to that prop.

PR #281, opened 2023-04-15
Update stub of sanctum in breeze to match config file of sanctum

maskeynihal:
1. Update the stub to match https://github.com/laravel/sanctum/blob/3.x/config/sanctum.php
2. Allow `port` in the URL for `FRONTEND_URL`

### DIFF
--- a/stubs/api/config/sanctum.php
+++ b/stubs/api/config/sanctum.php
@@ -1,5 +1,7 @@
 <?php
 
+use Laravel\Sanctum\Sanctum;
+
 return [
 
     /*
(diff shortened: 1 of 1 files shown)


taylorotwell:
Updated the two comments and gurad. If you want to change the URLs just tweak it in your own application.

maskeynihal:
@taylorotwell Should the config in breeze for `APP_URL` be `Sanctum::currentApplicationUrlWithPort()` instead of `env('APP_URL') ? ','.parse_url(env('APP_URL'), PHP_URL_HOST) : '',` like in [here](https://github.com/laravel/sanctum/blob/3.x/config/sanctum.php).

For the `FRONTEND_URL`, it doesn't match up with the example provided [here](https://github.com/laravel/breeze-next) as in the config it strips out `port`. I updated the config so that it works out of the box without changing anything by not stripping out the `port`. Or Should we update the documentation instead stating add  `SANCTUM_STATEFUL_DOMAINS` value should be added in the `.env` in this example?

PR #282, opened 2023-05-02
Add Test Cases for Negative Scenarios in User Registration

ysknsid25:
# Overview

Added validation for invalid passwords entered during user registration in the `test` and `pest-tests`.
The added patterns are:

- When the number of characters is less than the default 8 digits.
- When different passwords are entered for registration and confirmation.

# Reason

- So far, only positive test cases have been present in the user registration test, and I thought that negative test cases should also be tested.
- In the future development process, negative test cases may be inadvertently broken. To detect it in such cases.

That's all.
Thank you for your attention.

### DIFF
--- a/stubs/default/pest-tests/Feature/Auth/RegistrationTest.php
+++ b/stubs/default/pest-tests/Feature/Auth/RegistrationTest.php
@@ -19,3 +19,23 @@ test('new users can register', function () {
     $this->assertAuthenticated();
     $response->assertRedirect(RouteServiceProvider::HOME);
 });
+
+test('test new users cant register when password length less', function () {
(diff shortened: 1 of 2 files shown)


lazer-hybiz:
This is on the framework level, no need to test framework features in userland.

ysknsid25:
@lazer-hybiz 

No, this test is not limited to users only, but is also executed when contributing to Breeze. 

image

https://github.com/laravel/breeze/actions/runs/4858612129/jobs/8660227057

Therefore, I believe that it is important to perform negative testing to prevent code from breaking during the development process and ensure its stability.

taylorotwell:

PR #285, opened 2023-05-28
[1.x] Switch to `defineModel` in Vue stubs

osbre:
https://blog.vuejs.org/posts/vue-3-3

### DIFF
--- a/stubs/inertia-vue-ts/resources/js/Components/TextInput.vue
+++ b/stubs/inertia-vue-ts/resources/js/Components/TextInput.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import { onMounted, ref } from 'vue';
 
-const model = defineModel<string>({ required: true });
+const model = defineModel<string>({ required: true });
 
 const input = ref<HTMLInputElement | null>(null);
 
--- a/stubs/inertia-vue-ts/resources/js/Components/TextInput.vue
+++ b/stubs/inertia-vue-ts/resources/js/Components/TextInput.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import { onMounted, ref } from 'vue';
 
-const modelValue = defineModel();
+const modelValue = defineModel<string>({ required: true });
 
 const input = ref<HTMLInputElement | null>(null);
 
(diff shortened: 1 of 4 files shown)


jessarcher:
I'm excited about `defineModel` but I think we should keep this as a draft until the feature is no longer marked experimental in Vue.

selenofobico:
### **_How can I use the [ modal ] component in Breeze, any example of how to use it?_**

osbre:
Stable in 3.4: https://blog.vuejs.org/posts/vue-3-4
Will update the PR soon.

stubs/inertia-vue-ts/resources/js/Components/TextInput.vue
```diff
@@ -1,11 +1,7 @@
 <script setup lang="ts">
 import { onMounted, ref } from 'vue';
 
-defineProps<{
-    modelValue: string;
-}>();
-
-defineEmits(['update:modelValue']);
+const modelValue = defineModel();
```
Jubeki, line 4:
```suggestion
const modelValue = defineModel<string>();
```

I do not use Vue much, and this is untested and just from reading the docs: https://vuejs.org/api/sfc-script-setup.html#usage-with-typescript

stubs/inertia-vue/resources/js/Components/TextInput.vue
```diff
@@ -1,14 +1,7 @@
...
 
-defineProps({
-    modelValue: {
-        type: String,
-        required: true,
-    },
-});
-
-defineEmits(['update:modelValue']);
+const modelValue = defineModel();
```
Jubeki, line 4:
```suggestion
const modelValue = defineModel({
    type: String,
    required: true,
});
```
I do not use Vue much, and this is untested and just from reading the docs: https://vuejs.org/api/sfc-script-setup.html#definemodel

stubs/inertia-vue-ts/resources/js/Components/TextInput.vue
Jubeki: Maybe add here the required part to? Seems inconsistent with the non typescript part ?
```suggestion
const modelValue = defineModel<string>({ required: true });
```

jessarcher (approved):

jessarcher:
Thanks @osbre and @Jubeki!

PR #286, opened 2023-05-29
[1.x] Remove unnecessary union type on Vue refs

osbre:
From the Vue.js docs: https://vuejs.org/guide/typescript/composition-api.html#typing-ref

> If you specify a generic type argument but omit the initial value, the resulting type will be a union type that includes undefined:

> ```typescript
> // inferred type: Ref<number | undefined>
> const n = ref<number>()
> ```


jessarcher:
It's definitely nicer but goes against this section further down the page, specifically on typing template refs: https://vuejs.org/guide/typescript/composition-api.html#typing-template-refs.

osbre:
Thanks @jessarcher! You’re right I’ve missed that section 😅

PR #288, opened 2023-06-10
[Feature] Add. Support to Livewire Installation

ajmeireles:
### What?
This PR aims to introduce Livewire installation support.

### Why?
Since the fact that installing using the `blade` stack delivers both Tailwind and Alpine, with the introduction of Livewire developers will have easy access to the TALL Stack.

### Depends?
This PR depends on the PR created for [laravel/installer](https://github.com/laravel/installer/pull/264)


jessarcher:
I don't understand the point of just installing the Livewire package. Users can easily install that themselves if they want. If Breeze were to offer a Livewire option, I would expect the routes/views it creates for you to be Livewire components instead of plain Blade.

PR #291, opened 2023-06-16
Fix removing existing routes in the `web.php` file

mahmoudmohamedramadan:
When I was training on an Advanced Laravel project, I faced an issue with [laravel/ui](https://github.com/laravel/ui) so, I headed to use Laravel Breeze awesome package, and I had huge routes in the `web.php` file then after installation breeze blade I found that all my routes have been removed, and when I debugged the code I found that **Breeze** just copy the `web.php` of the package to the `routes` folder of my project not appended the code as the **laravel/ui** does, so I followed the **laravel/ui** approach of appending the routes and I have created a `web.stub` instead `web.php` then I append the content of stub file into the `web.php` file.

enjoy 🎉


driesvints:
Laravel Breeze is intended for new apps only, not existing ones sorry.

mahmoudmohamedramadan:
I think it's a false belief because the approach that I use is valid for any time installation.

askdkc:
@mahmoudmohamedramadan 

I love your way of doing this web.php file manipulation😉

This might be handy for many packages which intend to add some features, not just beginning of the project but when it needed.

Probably not suitable for starter kits like Breeze. But I really like your idea👍

mahmoudmohamedramadan:
@askdkc Thanks for your opinion.

PR #292, opened 2023-06-19
Use appName directly from .env in <title>

domnantas:
Currently, `appName` in app.tsx is pulled via DOM from app.blade.php <title> tag. It _works_, but using it from .env seems like a better approach. Additionally, it serves as an example of how .env variables should be accessed in React/Vue.

The `ssr` templates had `'Laravel'` hardcoded as the app name, which is not great for SEO. Replaced those parts with `import.meta.env.VITE_APP_NAME` as well.

This depends on https://github.com/laravel/laravel/pull/6204 where I added `VITE_APP_NAME` as an alias to `APP_NAME`

I don't think any changes in this PR could be considered breaking.

Similar issue fixed in Jetstream: https://github.com/laravel/jetstream/pull/1329

### DIFF
--- a/stubs/inertia-react-ts/resources/js/ssr.tsx
+++ b/stubs/inertia-react-ts/resources/js/ssr.tsx
@@ -4,7 +4,7 @@ import createServer from '@inertiajs/react/server';
 import { resolvePageComponent } from 'laravel-vite-plugin/inertia-helpers';
 import route from '../../vendor/tightenco/ziggy/dist/index.m';
 
-const appName = 'Laravel';
+const appName = import.meta.env.VITE_APP_NAME || 'Laravel';
 
 createServer((page) =>
     createInertiaApp({
--- a/stubs/inertia-react-ts/resources/js/app.tsx
+++ b/stubs/inertia-react-ts/resources/js/app.tsx
@@ -5,7 +5,7 @@ import { createRoot } from 'react-dom/client';
 import { createInertiaApp } from '@inertiajs/react';
 import { resolvePageComponent } from 'laravel-vite-plugin/inertia-helpers';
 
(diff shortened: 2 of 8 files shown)


stubs/inertia-react-ts/resources/js/ssr.tsx
```diff
@@ -4,7 +4,7 @@ import createServer from '@inertiajs/react/server';
 import { resolvePageComponent } from 'laravel-vite-plugin/inertia-helpers';
 import route from '../../vendor/tightenco/ziggy/dist/index.m';
 
-const appName = 'Laravel';
+const appName = import.meta.env.VITE_APP_NAME || 'Laravel';
```
domnantas, line 7: ssr can also use `process.env` to access envs (even without the `VITE_` prefix)

driesvints:
Tests fail here.

timacdonald (approved):

stubs/inertia-react-ts/resources/js/ssr.tsx
timacdonald: I like that it is consistent and uses `VITE_APP_NAME`, otherwise if you remove that ENV it could work in SSR and not a normal build.

timacdonald:
I've created a fix for the TS failure here: https://github.com/laravel/breeze/pull/293

It is unrelated to this PR.

jessarcher (approved):

stubs/inertia-react-ts/resources/js/ssr.tsx
domnantas: I agree. Using `process.env` would be a good example of accessing secrets/api keys and not exposing them to the frontend, but using it here would be a little misleading

PR #293, opened 2023-06-20
[1.x] Fix and standardise transitions

timacdonald:
We currently have a TS failure due to the Vue transition type not supporting the "class" attribute.

Migrated to the recommend attributes and also standardised the JSX version as well.

Vue Transition: https://vuejs.org/guide/built-ins/transition.html
Headless UI Transition: https://headlessui.com/react/transition

I've also remove the redundant `transform` class from transitions.

Screen Shot 2023-06-21 at 9 26 55 am


jessarcher (approved):

driesvints (approved):

domnantas:
It's always nice seeing TS catching a bug :)

PR #294, opened 2023-07-04
Auth controller regenerate session called twice

clementmas:
It looks like the Laravel framework is already regenerating the session automatically when logging in.

Here's the flow:
- AuthenticatedSessionController@store
    - `$request->authenticate()`
- LoginRequest@authenticate
    - `$this->attemptLogin()`
- Illuminate\Auth\SessionGuard@attempt
    - `$this->login($user, $remember)`
    - `$this->updateSession($user->getAuthIdentifier())`
    - `$this->session->migrate(true)`

So we don't need to manually call `$request->session()->regenerate()` (which then calls `$this->session->migrate()`).

We probably still need to manually call `$request->session()->regenerateToken()` though. I'll amend the PR based on your feedback.

### DIFF
--- a/stubs/default/app/Http/Controllers/Auth/AuthenticatedSessionController.php
+++ b/stubs/default/app/Http/Controllers/Auth/AuthenticatedSessionController.php
@@ -27,8 +27,6 @@ class AuthenticatedSessionController extends Controller
     {
         $request->authenticate();
 
-        $request->session()->regenerate();
-
         return redirect()->intended(RouteServiceProvider::HOME);
     }
 


taylorotwell:
Is this fixing any bugs in terms of user experience?

clementmas:
It doesn't create bugs on a default Laravel Breeze install. It's just confusing when building on top of it.

I couldn't figure out why the sessions id didn't match until I realized they were being regenerated twice.


I had assumed Laravel wouldn't 
In the end I replaced `$request->session()->regenerate()` with `$request->session()->regenerateToken()` and it behaves properly.

clementmas:
It doesn't create bugs on a default Laravel Breeze install. It's just confusing when building on top of it.

I couldn't figure out why the session id before `regenerate()` didn't match the old session id until I realized it was being regenerated twice.
It's not really efficient/clean. I've been spoiled with Laravel being so well built so I assumed it was my fault.

In the end I replaced `$request->session()->regenerate()` with `$request->session()->regenerateToken()` and it behaves properly.

zawilliams:
I actually ended up having an issue with this exact double regeneration. I'm dealing with a legacy system and am listening for the `Login` event and then logging in the legacy system after using Laravel's `session()` helper and the double regeneration was causing the legacy system to not be logged in. 

Removing the `$request->session()->regenerate()` does seem to have fixed things in terms of this being a bug.

PR #300, opened 2023-08-08
[1.x] Pest detection

jessarcher:
This PR updates the standalone Breeze installer to default to Pest when it is already installed, either via the Laravel installer or installed manually.

When running `php artisan breeze:install` with no arguments, the testing framework prompt will still be displayed, but defaulting to Pest when installed:

image

When passing arguments to the installer (e.g. `php artisan breeze:install blade`) the prompt does not appear. Currently, it will install PHPUnit tests unless the `--pest` flag was passed. With this change, it will automatically install Pest tests when appropriate.

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -73,8 +73,10 @@ class InstallCommand extends Command implements PromptsForMissingInput
 
         $stubStack = $this->argument('stack') === 'api' ? 'api' : 'default';
 
-        if ($this->option('pest')) {
-            $this->removeComposerPackages(['phpunit/phpunit'], true);
+        if ($this->option('pest') || $this->isUsingPest()) {
+            if ($this->hasComposerPackage('phpunit/phpunit')) {
+                $this->removeComposerPackages(['phpunit/phpunit'], true);
+            }
 
             if (! $this->requireComposerPackages(['pestphp/pest:^2.0', 'pestphp/pest-plugin-laravel:^2.0'], true)) {
                 return false;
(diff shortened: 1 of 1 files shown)


src/Console/InstallCommand.php
```diff
@@ -73,8 +73,10 @@ protected function installTests()
 
         $stubStack = $this->argument('stack') === 'api' ? 'api' : 'default';
 
-        if ($this->option('pest')) {
-            $this->removeComposerPackages(['phpunit/phpunit'], true);
+        if ($this->option('pest') || $this->isUsingPest()) {
+            if ($this->hasComposerPackage('phpunit/phpunit')) {
+                $this->removeComposerPackages(['phpunit/phpunit'], true);
+            }
```
jessarcher, line 79: This solves an existing minor issue when the user already has Pest. Previously the console output would show a warning about the package not existing, and it would unnecessarily run the composer scripts, which takes a bit of time.

jessarcher:
Ultimately I think it would be nice to never prompt for the testing framework outside of the Laravel installer, and just autodetect the appropriate testing framework in the installers and Artisan `make` commands, instead of needing `--pest` flags, etc.

driesvints (approved):

nunomaduro (approved):
@jessarcher I've improved the `isUsingPest` function.

jessarcher:
@nunomaduro Couldn't that class exist just because of another dependency?

I figured the safest way was to check whether the user had any Pest-specific files in their own project.

PR #301, opened 2023-08-08
Fix 404 error on password reset link received in emails

uaibo:
Password reset link was wrong and it would lead to 404 error.

the link sent to email was this: 
http://localhost:3000/password-reset/{TOKEN}?email={EMAIL}

and it should have been this:
http://localhost:3000/reset-password/{TOKEN}?email={EMAIL}

by using named route we make sure the correct link is sent and this problem doesn't happen again.


taylorotwell:
Meant to be paired with this, which uses the name in the stub: 

https://github.com/laravel/breeze-next/tree/master/src

PR #303, opened 2023-08-29
replace in file if the file exists

saeedvaziry:
Currently when you try to install Breeze with Blade and you don't have the welcome.blade.php file it fails because it is not able to find the welcome.blade.php


saeedvaziry:
Closing as it might not be a good use case

PR #309, opened 2023-09-05
sending the verification email to the user before redirecting to the …

hdood:
The bug is that when redirecting the user to verify-email.blade.php view we don't send a verification email to him, but we display a button with the text RESEND VERIFICATION EMAIL which means that the email was already sent but it's not in EmailVerificationPromptController we just check email verification and redirect the user to verify-email.blade.php without sending the verification email.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 composer.lock
 /phpunit.xml
 .phpunit.result.cache
+.vscode
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,18 @@
+{
+    "workbench.colorCustomizations": {
+        "activityBar.background": "#8155BA",
+        "activityBar.foreground": "#281C2D",
+        "activityBar.inactiveForeground": "#252525",
+        "activityBar.activeBorder": "#7eaa45",
+        "activityBar.activeBackground": "#695E93",
+        "activityBar.border": "#695E93",
+        "titleBar.activeBackground": "#8155BA",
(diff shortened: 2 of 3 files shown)


jessarcher:
Hi @hdood,

The initial verification email should be sent when the user registers, as long as the `User` model implements the `Illuminate\Contracts\Auth\MustVerifyEmail` contract. It shouldn't be resent every time they try to access a page.

See: 
* https://github.com/laravel/laravel/blob/32ecad53a9a59d7ad33dbfd6352a5ea0fd0da2cf/app/Providers/EventServiceProvider.php#L18-L20
* https://laravel.com/docs/10.x/verification#model-preparation

Also, it's generally [bad practice](https://www.rfc-editor.org/rfc/rfc2616#section-9.1) for a `GET` request to have side effects, like sending an email.

.vscode/settings.json
jessarcher, line 1: This is editor-specific code, not package-specific code.

.gitignore
```diff
@@ -2,3 +2,4 @@
 composer.lock
 /phpunit.xml
 .phpunit.result.cache
+.vscode
```
jessarcher, line 5: Editor-specific ignores belong in your global gitignore. This is just for package-specific ignores.

hdood:
Sorry for the errors this is my first pull request, Thank you so much Jess for the references.

jessarcher:
No worries, @hdood! Thanks for submitting a PR! :slightly_smiling_face:

PR #310, opened 2023-09-06
[1.x] Add test in case the email was already verified.

xiCO2k:
This PR adds missing test in case the email was already verified.


taylorotwell:
Going to table this one for now.

PR #313, opened 2023-09-14
Compatibility with livewire 3

LeMatosDeFuk:
I've found an old compatibility with livewire 3, where alpine is already included in core files.


driesvints:
Thanks but we're currently already working on a PR for this. We'll hopefully have something concrete soon 👍

PR #314, opened 2023-09-15
[1.x] Adds Livewire stack 🐙

nunomaduro:
<p align="center">
    breeze:install
</p>

This pull request introduces the Livewire 🐙 stack to Laravel Breeze. It uses Volt for single file components, and its based class API. Here's how you can test this pull request locally:

```bash
# Select "No starter kit"
laravel new breeze-livewire

cd breeze-livewire
composer require laravel/breeze:dev-feat/livewire
php artisan breeze:install livewire --dark --pest
sed -i.bak '/DB_/d' ./.env
echo "DB_CONNECTION=sqlite" >> .env
php artisan migrate --force
php artisan serve
```

The functional API will follow, in a separate pull request.


src/Console/InstallsLivewireStack.php
```diff
@@ -0,0 +1,96 @@
...
+            return [
+                '@tailwindcss/forms' => '^0.5.2',
+                'autoprefixer' => '^10.4.2',
+                'postcss' => '^8.4.6',
+                'tailwindcss' => '^3.1.0',
+            ] + $packages;
+        });
+
+        // Install Livewire...
+        if (! $this->requireComposerPackages(['livewire/livewire:dev-main as 3.0.4', 'livewire/volt:^1.0'])) {
```
driesvints, line 29: Do we need to require Livewire if we already require Volt? Volt already brings in Livewire.

nunomaduro: No. However, I need that line temporarily for now.

lucasjose501:
While I have some reservations about using Volt due to my personal coding preferences, I must commend the incredible speed from Livewire and the thoughtful implementation here! 🚀🚀

lucasjose501:
> I don't have anything against Volt, but would be great to have an option to use Volt or plain Livewire components. Is that something that's planned?

2 versions maybe? "Livewire with Volt" and "Livewire with default components"

Jubeki:
Works great!

It seems like the Livewire stack is missing from the test matrix.

nunomaduro:
@Jubeki well spotted!

PR #315, opened 2023-09-24
fix: :bug: bring navigation links on top of other elements

MooseSaeed:
## The Problem
When you install a new Laravel Breeze project, you'll notice that on desktop screens, the navigation links can sometimes get buried beneath other content, like cards.

## Examples:
laravel-breeze-nav-bug

Screenshot of Breeze's navlinks issue

## The Solution
This change is to ensure that the sticky navbar in Laravel Breeze always stays on top of the page's elements which is the expected behavior of any navbar.

## After
laravel-breeze-after-fixing-navlinks


taylorotwell:
I think it's OK.

PR #316, opened 2023-09-25
Removed PHPStan errors which would show up after installing the breeze package.

romek-dev:
I've recently started multiple projects in a row, and as always i instantly download Larastan and set the Level to 9.
When using the breeze package, i have to manually go through all the files and fix the PHPStan errors myself.

This PR fixes this issue and removes the PHPStan errors. I think this will benefit many developers that also use PHPStan/Larastan by default, including myself.

If i made any mistakes in the code / assumed the wrong types or anything else, tell me and I will gladly iterate over this PR and correct the mistakes.


taylorotwell:
Sorry a lot of this just feels like noise.

romek-dev:
> Sorry a lot of this just feels like noise.

Hi, if you have the time, could you elaborate on this?
Do you not think that this would greatly benefit some developers that use PHPStan by default?
Of course, this is your decision to make, but i would appreciate it if you could tell me your reasoning.

PR #319, opened 2023-10-10
Lowercase email on registration

taylorotwell:
Similar to: https://github.com/laravel/fortify/pull/485

Postgres is case sensitive when comparing strings. If someone is using Breeze with Postgres they could have users register accounts with "taylor@laravel.com" **and** "Taylor@Laravel.com" for example without any errors.

This fixes that by lowercasing emails on registration like Fortify / Jetstream now does by default.


nunomaduro:
Duplicated of https://github.com/laravel/breeze/pull/321.

PR #333, opened 2023-12-04
Dispatching a component event to itself

akr4m:


akr4m:
After updating profile information & password,  it will dispatch event to itself to show `action` message.

[https://livewire.laravel.com/docs/events#dispatching-events-from-blade-templates](https://livewire.laravel.com/docs/events#dispatching-events-from-blade-templates)

taylorotwell:
No info about what this is fixing?

PR #335, opened 2023-12-06
[1.x] Change Attribute of Authentication Exception

devajmeireles:
Related to: https://github.com/tallstackui/tallstackui/issues/116

When the `authenticate()` throws `ValidationException` it should link the error with `form.email` instead of `email` because it is defined in Livewire Forms. With that, Blade validations based on `form.email` will work successfully.

### DIFF
--- a/stubs/livewire-common/app/Livewire/Forms/LoginForm.php
+++ b/stubs/livewire-common/app/Livewire/Forms/LoginForm.php
@@ -34,7 +34,7 @@ class LoginForm extends Form
             RateLimiter::hit($this->throttleKey());
 
             throw ValidationException::withMessages([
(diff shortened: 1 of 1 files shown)


taylorotwell:
Made this change locally and no longer see the error message in my Breeze app.

devajmeireles:
> Made this change locally and no longer see the error message in my Breeze app.

It's not me saying it, it's Livewire itself. Review the example of the validation error in Blade component from this link:
https://livewire.laravel.com/docs/forms#extracting-a-form-object

ChrisRiddell:
Seems `login.blade.php` will also need to be updated to reflect this change and to match the Livewire docs.

PR #344, opened 2024-01-02
change require to require_once on web.php

anasmorahhib:
will check if the file has already been included.


driesvints:
Thanks but this isn't necessary.

PR #346, opened 2024-01-13
Add client side hydration to react stubs

matthiasweiss:
The "Client side hydration" section in the SSR documentation of Inertia (https://inertiajs.com/server-side-rendering#client-side-hydration) mentions the following for React:

> "To enable client-side hydration in a React app, update your app.js file to use hydrateRoot instead of createRoot [..]"

This is currently not reflected in the React stubs for Breeze, they ship with `createRoot` by default.

This PR changes the calls in the JS and TS stubs. It also adds another `@ts-expect-error` statement to hide a type error related to accessing the route on the `global` object.


matthiasweiss:
Wrong base branch 😢  Will resubmit asap

PR #347, opened 2024-01-13
Enable client site hydration in React stubs

matthiasweiss:
The "Client side hydration" section in the SSR documentation of Inertia (https://inertiajs.com/server-side-rendering#client-side-hydration) mentions the following for React:

grafik

This is currently not reflected in the React stubs for Breeze, they ship with `createRoot` by default.

This PR changes the calls in the JS and TS stubs.


taylorotwell:
Will leave this user opt-in for now.

PR #348, opened 2024-01-14
Add default value for useRef calls in .tsx files

matthiasweiss:
This fixes missing default values for `useRef` calls that use generics, in this case `HTMLInputElement`.

Currently the returned value is inferred as `HTMLInputElement | null | undefined`. Adding null as a default value is better, since the returned value is inferred as `HTMLInputElement | null` due to the signature `function useRef<T>(initialValue: T | null): RefObject<T>;`.

I ran into this issue because I swapped the `<TextInput />` in the `<UpdatePasswordForm />` component with a different input component (https://ui.shadcn.com/docs/components/input), which caused an error when passing the `ref` down to the new component. Adding null as the default value fixed that issue.


jessarcher (approved):
Looks good to me. The React docs demonstrate passing an initial value of `null` in this scenario too.

https://react.dev/reference/react/useRef#manipulating-the-dom-with-a-ref

PR #349, opened 2024-01-19
Add Livewire with components option

dougffer:
This PR adds the option to install Livewire with components, without Volt. It renames old livewire stubs folders and adds the `livewire-components` option to breeze:install.

I really love Livewire, but i couldn't get used to Volt's code style, so i decided to create this to offer a new option in Breeze for other developers who also prefer livewire Without Volt.

I copied most of the codes from the volt stack and the Blade stack to create Livewire components and blades.

Example:
```shell
$ php artisan breeze:install livewire-components
``` 

In stubs, i decided to rename the `livewire` and `livewire-common` folders to avoid confusion for other developers and renamed `livewire-functional` to ensure consistency with folder names. So, the new structure with new option is:

- ...
- stubs/livewire-components
- stubs/livewire-volt-class
- stubs/livewire-volt-common
- stubs/livewire-volt-functional

I also needed to rename the trait that installs the stack, from `InstallsLivewireStack` to `InstallsLivewireVoltStack`, and adds `InstallsLivewireComponentsStack`.

For some reason, GitHub is confused about changes in folder names when there is an identical file in the new folder `livewire-components`. However, in the first commit, the comparison of folders is correct.

I hope you like it, thanks.


taylorotwell:
I don't think we want to take on another stack to maintain.

PR #355, opened 2024-02-08
Add Authentication By Use Of API Tokens (Sanctum)

kemboi22:
- Added Authentication by use of Sanctum Tokens 
- The Response is returned of json 
- Authentication of APIs using Sanctum  sessions is dissabled


driesvints:
Tests fail here

PR #357, opened 2024-02-14
Support distinct PK for `User` model

GromNaN:
Some projects use a distinct primary key for the `User` model. This change makes sure the validation constraint is checking the correct field with the correct value defined by the model class.
This affect the "Profile Information" form with the email field.


taylorotwell:
You can change this in your own app.

GromNaN:
Of course, like I could create everything in my own app. The goal is to make it work out of the box for other Laravel users, with minimal changes.

PR #361, opened 2024-02-27
Controllers were not being copied when using livewire as setup guide.

immuhammad:
When using breeze package with Livewire as a setup guide, the controllers were not being copied. 

Only the `stubs/default/app/Http/Controllers/Auth/VerifyEmailController.php` to `Http/Controllers/Auth/VerifyEmailController.php`  but rest of the files in controller were being ignored resulting in breeze package not working as it should be.


taylorotwell:
I just installed Breeze today and it was working fine?

PR #369, opened 2024-03-14
Add livewire stack with no volt

sohailahmad07:
Hi 

i added a new stack to Laravel Breeze that uses just livewire without volt i also did testing locally and installed breeze
all three stack of livewire with both PHPUnit and Pest and everything seems to be an order

before all of the livewire volt class API files were in livewire i moved them to livewire-volt-class-api

File Moved From: livewire -> livewire-volt-class-api

and in the livewire directory, i add files for only the livewire stack NO VOLT


taylorotwell:
No plans to do this atm.

PR #370, opened 2024-03-14
Added Authentication Using Sanctum Api Tokens

kemboi22:
Authentication Api Tokens Sanctum Api Tokens for people who dont use sessions to authenticate their users


kemboi22:
reason for refusal for this

PR #371, opened 2024-03-15
Adds a check for the existence of the `resources/views` directory before adding `.gitkeep`

MrShnaider:
When developing an api application, directories `resources/views` or even `resources` may be deleted before breeze installation. When installing breeze with the `api` option, copying stub files to such projects breaks down.

PR adds a simple check for the existence of such directories before creating the `resources/views/.gitkeep` file.

[Similar issue](https://github.com/laravel/framework/issues/50586)


MrShnaider:
Write at least the reason for the refusal, please

driesvints:
@MrShnaider Breeze is only for new projects, not existing ones.

PR #374, opened 2024-03-21
[2.x] Prevents remove `dark` classes from `livewire/welcome/navigation.blade.php`

devajmeireles:
This pull request aims to add `notPath` to Finder to avoid removing `dark` classes from `resources/views/livewire/welcome/navigation.blade.php`. When `dark` is removed from this file the Login and Registration links become invisible:

CleanShot 2024-03-21 at 14 01 04

CleanShot 2024-03-21 at 14 01 50

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 composer.lock
 /phpunit.xml
 .phpunit.result.cache
+.idea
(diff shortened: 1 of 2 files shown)


nunomaduro:
Thanks! Almost perfect. Created https://github.com/laravel/breeze/pull/375, because was missing fixing the functional stack too.

devajmeireles:
> Thanks! Almost perfect. Created #375, because was missing fixing the functional stack too.

You could have mentioned that, I would make the necessary adjustment.

PR #376, opened 2024-03-30
Passing hint in multiselect.

lmottasin:
It is hard for a beginner or anyone without knowledge about console applications to find out how to select in multi-select. 
For example, he needs something from optional like dark more, ssr but doesn't know how to select that, press enter, and shows nothing is selected. So, for an easier way to install and better UX, there should be a hint, on how to select in multi-select.
No breaking change just passed a hint string which is also a default string in helper.


taylorotwell:

PR #378, opened 2024-04-07
Use post named routes inside of blade files

aniket-magadum:
Currently, we are using named routes for both register and login, but it is pointed to the get method.

But in case we want to make any changes to the route path for example as below we get a 405 method not allowed exception

```php
    Route::get('admin/login', [AuthenticatedSessionController::class, 'create'])
        ->name('login');
```

image

I have instead provided a name to the post route and used them in the blade forms.


aniket-magadum:
Not sure if the failing tests are related to this PR

taylorotwell:
Feel free to change in your app. 👍

PR #379, opened 2024-04-17
Refactor email verification tests

ghost:
Currently, the email verification tests do not use the User factory's `unverified` helper. They instead override the `email_verified_at` attribute's default value to `null` via the `make` method. This PR refactors said tests to make use of the `unverified` helper.


javadihugo:
Wait, is it possible someone might first install an old version of Laravel where the installer doesn't generate `unverified` as part of the boilerplate?

PR #389, opened 2024-06-06
Bump `@headlessui/react` to v2

RobinMalfait:
This PR bumps `@headlessui/react` to the latest v2 version.

There are 3 relevant changes at play here:

1. Bump the actual `@headlessui/react` version.
2. Remove the `as={Fragment}` prop from the `<Transition>` and `<Transition.Child>` components. This is now the default behavior.
3. Replace the deprecated dot notation with concrete imported components.

These changes are optional because they work in a backwards compatible way. But since the `as={Fragment}` prop is now the default behavior, it's cleaner to remove it.

The dot notation will also still work, but since it's deprecated, your editor will show a warning, so let's clean it up as well.

Apart from these steps nothing else is changed nor required.

### DIFF
--- a/src/Console/InstallsInertiaStacks.php
+++ b/src/Console/InstallsInertiaStacks.php
@@ -190,7 +190,7 @@ trait InstallsInertiaStacks
         // NPM Packages...
         $this->updateNodePackages(function ($packages) {
             return [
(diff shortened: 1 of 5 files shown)


taylorotwell:
Thanks!

PR #390, opened 2024-06-24
Remove Hash Facade from Authentication Controllers

ikamal7:
This PR aims to remove the `Hash` facade from the authentication controllers, specifically where `Hash::make($request->string('password'))` is used for password hashing. This change is made to avoid duplicate password hashing, as it is already handled in the `User` model. By doing so, we can ensure consistency and reduce redundancy in our codebase.

### Reasoning:

- **Consistency:** Password hashing should be managed in one place, ensuring that any changes to the hashing mechanism are applied universally.
- **Redundancy:** Removing duplicate hashing logic reduces redundancy and potential errors.
- **Maintenance:** Centralizing password hashing in the `User` model simplifies maintenance and future updates to the hashing process.

### Implementation:

- Changed from `Hash::make($request->string('password'))` to `$request->password`.

Controllers Modified:
- `RegisteredUserController` (api, default and inertia  )
- `NewPasswordController` (default and inertia)

### DIFF
--- a/stubs/api/app/Http/Controllers/Auth/RegisteredUserController.php
+++ b/stubs/api/app/Http/Controllers/Auth/RegisteredUserController.php
@@ -8,7 +8,6 @@ use Illuminate\Auth\Events\Registered;
 use Illuminate\Http\Request;
 use Illuminate\Http\Response;
 use Illuminate\Support\Facades\Auth;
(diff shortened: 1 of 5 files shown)


taylorotwell:
No plans to change this at this time.

PR #392, opened 2024-07-03
Remove `disabled` prop

MartijnCuppens:
I started learning Laravel lately, so I had a look at you guys made the blade components. I noticed `disabled` was an explicit prop, while others like `required` or `autofocus` are not. I think this can be removed, unless there is some other magic going on I'm not aware of yet.


TENIOS:
Please tell me why the use of Blade [`{!! !!}`](https://laravel.com/docs/11.x/blade#displaying-unescaped-data) statements instead of `{{ }}` like in all other components?

https://github.com/laravel/breeze/blob/8a6032767088e3cc357e7014ecdb62d2594a3fa8/stubs/default/resources/views/components/text-input.blade.php#L3

Thanks in advance for reply.

PR #407, opened 2024-09-08
Add Extra Tests

RobHarveyDev:
Adds some extra tests for the default blade scaffolding, so that the tests have 100% coverage.


taylorotwell:

PR #412, opened 2024-09-16
[2.x] Add option to install ESLint with Prettier

avosalmon:
CleanShot 2024-09-12 at 09 55 53

This PR introduces an option to install and configure ESLint with Prettier for the Inertia stacks.
Selecting the `ESLint with Prettier` option will install ESLint and Prettier with some relevant plugins and add config files.

- It still uses ESLint 8 because some plugins don’t support ESLint 9 flat config yet.
  - e.g. https://github.com/facebook/react/issues/28313
- It adds the `lint` node script to package.json
- This PR also adds a CI workflow to run Pint and ESLint and then commit the coding style changes.

> [!NOTE]  
> Most of the changes for the stubs are code style changes by CI (14759fc864a37a7374de23c88af4336910bb7a25, 60a5a2a48e9ba8df8628c283b448a35344c645dd). Following the commit logs would make it easier to review the changes rather than diving into the files changed.

PR #419, opened 2024-09-26
[2.x] Add option to install Biome with React Inertia stack 

JoeyMckenzie:
image

Tailing off @avosalmon's PR for adding [ESLint and Prettier](https://github.com/laravel/breeze/pull/412), I think it would be handy to also include [Biome](https://biomejs.dev/) as an option for the React Inertia stack. This option would be mutually exclusive with the ESLint/Prettier option as they serve the same purpose. I've been using Biome in some recent projects and am quite fond of it, often preferring it to ESLint/Prettier. It would be great to offer it as an option for those picking React without the need to manually integrate it into the template. 

A key few highlights:
- Includes a configuration preset that I've found helpful when using in my own projects (ignores `ssr.tsx`/`ssr.jsx` as there's some cryptic errors in there that I haven't fully figured out the root cause of just yet 🫠)
- Adds some npm scripts to format and lint JS/TS files based on Biome's [usage docs](https://biomejs.dev/guides/getting-started/#usage)

Outstanding items/points of discussion:
- Running the linter will introduce a few minor errors around button semantics (adding a11y options to the JSX/TSX), hook deps, and some SVG warnings
- I've opted to leave out the fixed code for now, though am more than happy to provide stubs if the general consensus on DX is to provide a correctly linted scaffolding out of the box specifically for the Biome option 

Thanks to @avosalmon for introducing ESLint/Prettier initially, been waiting for that for a long time!


taylorotwell:
Don't really want to maintain multiple stacks for this. 😕

PR #423, opened 2024-10-02
Always prompt for optional features

reibengu:
Started a new project using Laravel Bootcamp and ran `php artisan breeze:install react` as instructed.

Noticed the installer doesn't prompt for optional features like TypeScript support or Dark mode when stack is set.
It led to confusion and forced me to recreate the project to include these features.

Changed logic to always prompt for optional features and select as default the arguments passed directly like  `php artisan breeze:install react --typescript`.

It might require an extra argument `--no-interaction` for cases when this is run from another script, similar to what composer uses (that can also be a global trait of commands).

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -51,6 +51,8 @@ class InstallCommand extends Command implements PromptsForMissingInput
      */
     public function handle()
     {
+        $this->promptForOptionalFeatures();
+
         if ($this->argument('stack') === 'vue') {
             return $this->installInertiaVueStack();
         } elseif ($this->argument('stack') === 'react') {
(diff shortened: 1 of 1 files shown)


reibengu:
I can add `--no-interaction` to make the tests pass in case you approve the idea.

crynobone:
This break the current expected behaviour with PromptsForMissingArgument contract. You should provide `breeze:install` instead of providing any argument.

reibengu:
> This break the current expected behaviour with PromptsForMissingArgument contract. You should provide `breeze:install` instead of providing any argument.

It's not related to `PromptsForMissingInput` at all, it's related to how the CLI is designed and if it should always ask for the optional arguments or not.

taylorotwell:
We actually don't document passing arguments to this command.

PR #426, opened 2024-10-04
[2.x] Fixed missing alt attribute in IMG tag

amdad121:
Fixed missing alt attribute in IMG tag in Welcome template.


taylorotwell:
This page is never meant to be shown to end users.

PR #429, opened 2024-10-11
Update Login/Register form to look more modern

mezotv:
This makes a few changes to the login and register form including the following:

- "Don't have an account" yet button on login
- Make the action button full width
- Add a heading to both pages

This aims to make both forms look more modern and easier on the eye. 

Preview: 

Bildschirmfoto 2024-10-11 um 12 25 20
Bildschirmfoto 2024-10-11 um 12 25 08


mezotv:
As you may have noticed this is not added to all the templates yet, I want to make sure this layout is accepted before and then I will add the rest of them to this pull request! So please let me know @taylorotwell!

taylorotwell:
Hey there - we will be working on the starter kits in general soon so we'll take a look at this then!

PR #430, opened 2024-10-17
[2.x] Fixed TypeScript Errors in React SSR Setup with Ziggy Props when Builds

akr4m:
This PR fixes TypeScript errors encountered during the SSR build process for React. Specifically, two errors were addressed:
1. `Spread types may only be created from object types` due to `page.props.ziggy` being of type `unknown`.
2. TypeScript type issue when trying to access `page.props.ziggy.location`.

The updated code includes `@ts-expect-error` directives to bypass the type errors temporarily and allow successful builds.

A screenshot of the errors is attached for reference.
image


akr4m:
I just tested it again.
The error is showing when I'm using `pnpm`.

Using `npm` if fine.

PR #431, opened 2024-10-23
fix(sanctum): add 127.0.0.1:3000 for proper API auth during local development with 127.0.0.1:3000

shahshaikat:
fix(sanctum): add 127.0.0.1:3000 for proper API auth during local development with 127.0.0.1:3000


shahshaikat:
After facing this issue multiple time just found a solution of a common issue with 127.0.0.1:3000 wasn't working for a simple bug.

PR #433, opened 2024-10-29
[2.x] Add frontend_url to App Config in API Stack

amdad121:
Some files in the project use app.frontend_url (e.g., [this search result](https://github.com/search?q=repo%3Alaravel%2Fbreeze+app.frontend_url&type=code)), but it was missing from the config/app.php file in the API stack. I have added the feature by including the frontend_url in the app config.


taylorotwell:
You can add it to your app if needed.

PR #434, opened 2024-11-02
fix(ts): remove redundant @ts-expect-error directives in SSR route setup

datlechin:
The `@ts-expect-error` directives were unnecessary for the Ziggy props and location assignments within the global route function, as TypeScript can properly infer these types.

Screenshot 2024-11-02 at 12 38 36


reyberyturiaga:
To add a bit of context, this issue was introduced by https://github.com/laravel/breeze/pull/430. But that PR is failing the automated check for the thing it's trying to fix: https://github.com/laravel/breeze/actions/runs/11384357740/job/31671889124. And indeed, on a fresh install of Laravel Breeze with `--react` and `--typescript` flags, the build is failing. Removing `@ts-expect-error` directives make the build work again.

PR #446, opened 2025-01-17
[2.x] add types to `useForm` in `Login.tsx`

joelbutcher:
This PR adds type definitions to the `useForm` props in `Login.tsx` for the React Typescript stack, which fixes the problem with [these recent tests failing](https://github.com/laravel/breeze/actions/workflows/tests.yml?query=branch%3A2.x)


taylorotwell:
Hey there! We are rebuilding our starter kits now, so will address any needs there. We should have them out in ~4 weeks.

joelbutcher:
Hey @taylorotwell I appreciate that you and the team are shifting focus for starter kits, but in the meantime this is currently causing my test pipelines to fail for Socialstream as I try as best I can to follow the files that are published by Laravel (to avoid making too many changes if someone installs Breeze, commit it, then add Socialstream later in a different commit).

If this is a hard "no", I'll just override this in my packages stub files to get my tests passing 👍

PR #448, opened 2025-01-21
[2.x] Add pagination type definitions

chikondikamwendo:
This PR adds pagination type definitions to both React and Vue Typescript flavors for Inertia. It aims to simplify how pagination objects are imported into Inertia projects from the backend.

Example in React:
```ts
// imports...

export default function Todos({ pagination }: PageProps<{ pagination: Pagination<Todo> }>) {
    return (
        <>
            {/* ... */}
        </>
    );
}
```


taylorotwell:
We are rebuilding our starter kits now. More info soon.

PR #450, opened 2025-01-22
[2.x] Give users the option to keep js/css when installing the API stack

paulrose:
Give users the option to keep js/css when installing the API stack.

For example, it's often a case where I'll want to add Auth features, but I've already started working on my frontend, I just don't want to implement the full Breeze stack for whatever reason.


taylorotwell:

paulrose:
> Thanks for your pull request to Laravel!
> 
> Unfortunately, I'm going to delay merging this code for now. To preserve our ability to adequately maintain the framework, we need to be very careful regarding the amount of code we include.
> 
> If applicable, please consider releasing your code as a package so that the community can still take advantage of your contributions!

@taylorotwell, While I appreciate what you're saying, you need to be careful about the amount of code you introduce, I'd ask you to re-think the approach here. 

I'm unsure why adding authentication routes would indiscriminately delete a users project files without warning? That feels like bad design. File deletion is not mentioned in the docs at all.

PR #453, opened 2025-01-23
fix: resolve type error in 'setData' for 'remember' parameter

koficoud:
Fixed a type error in Login.tsx (line 81) where setData was being called with a boolean value (e.target.checked) for the remember parameter, which expected a value of type false.

Solution: Adjusted the logic or type definition to ensure compatibility between the value passed (boolean) and the expected type for the remember parameter.

`
resources/js/Pages/Auth/Login.tsx:81:53 - error TS2345: Argument of type 'boolean' is not assignable to parameter of type 'false'.
`

81                                 setData('remember', e.target.checked)`

### DIFF
--- a/stubs/inertia-react-ts/resources/js/Pages/Auth/Login.tsx
+++ b/stubs/inertia-react-ts/resources/js/Pages/Auth/Login.tsx
@@ -17,7 +17,7 @@ export default function Login({
     const { data, setData, post, processing, errors, reset } = useForm({
         email: '',
         password: '',
-        remember: false,
+        remember: false as boolean,
     });
 
     const submit: FormEventHandler = (e) => {


taylorotwell:
Where would remember always expect false? Something about this doesn't seem correct. 🤔

koficoud:
> Where would remember always expect false? Something about this doesn't seem correct. 🤔

Hi Taylor,

Thank you for your feedback and for taking the time to review my PR. I understand that something about the logic for the remember parameter didn't seem correct. Could you clarify if the parameter is intended to strictly expect false, or if it should support boolean values (true or false)?

I’d be happy to adjust my implementation to align with the intended behavior and resubmit a more accurate fix.

Thanks again for the opportunity to contribute to the project. 😊

PR #463, opened 2025-02-05
[2.x] [API stack] Add `APP_URL` and `FRONTEND_URL` to `.env.example`

josephmiclaus:
Please review #466 first.

Updates `.env.example` with `APP_URL=http://localhost:8000` and `FRONTEND_URL=http://localhost:3000`, in addition to the changes made to `.env` when installing the API stack.

Since the `.env` file is typically not committed, this ensures that `.env.example` contains the necessary environment variables for consistency when setting up the project.

Note: The install script already assumes that `.env.example` exists.


taylorotwell:
We are rebuilding our starter kits now.

PR #465, opened 2025-02-05
[2.x] [API stack] Remove `VITE_APP_NAME` from `.env` and `.env.example`

josephmiclaus:
Please review #466 first.

Removes the `VITE_APP_NAME="${APP_NAME}"` line from both the `.env` and `.env.example` files, since Vite not used in the API stack.

This PR depends on #463.


taylorotwell:
I think it's fine.

PR #466, opened 2025-02-05
[2.x] [API stack] Improve installation for Sail projects and ensure consistency in env files

josephmiclaus:
Improves the API stack installation considering Sail projects and ensures consistency between `.env` and `.env.example`.

### Consistency between `.env` and `.env.example`

Changes that were previously applied only to `.env` are now also applied to `.env.example`. This ensures that the `.env.example` file contains all the necessary environment variables, making it easier for team developers to set up the project. (Especially important for Sail projects, see below.)

### Considering API stack

**Remove `VITE_PORT` from `docker-compose.yml` and `VITE_APP_NAME` from environment files**: Since Vite is not used in the API stack, these are not needed.

### Considering Sail projects

**Add `APP_PORT` to environment files**: This is needed because when running `sail up`, the API is only accessible at http://localhost:80 instead of the intended http://localhost:8000. When `APP_PORT` is not set in the `.env` file, the [sail binary defaults it to port 80](https://github.com/laravel/sail/blob/1.x/bin/sail#L142), causing this issue. Modifying the ports in `docker-compose.yml` doesn't resolve this issue (as attempted in #464), because the `sail` binary exports `APP_PORT` before `docker-compose.yml` is applied. Adding `APP_PORT` to `.env` resolves the issue.

---
This PR combines the changes from #463 + #465, since they were interrelated.

### DIFF
--- a/src/Console/InstallsApiStack.php
+++ b/src/Console/InstallsApiStack.php
@@ -68,6 +68,9 @@ trait InstallsApiStack
             $envContent = file_get_contents($envPath);
             if (! empty($envContent)) {
                 $envVars = 'APP_URL=http://localhost:8000'.PHP_EOL.'FRONTEND_URL=http://localhost:3000';
+                if ($isSail) {
+                    $envVars = 'APP_PORT=8000'.PHP_EOL.$envVars;
+                }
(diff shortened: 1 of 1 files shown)


taylorotwell:
We are rebuilding our starter kits now.

PR #482, opened 2026-04-06
[Feature]: Inertia 3

joaopalopes24:
Update Breeze’s Inertia stacks (Vue and React) to Inertia v3 and `@inertiajs/vite`.

- PHP/JS dependencies aligned with v3; React 19 where required
- Vite uses the i`nertia()` plugin; client entries match the docs
- Optional SSR via _ssr.*_, Ziggy, and an explicit entry in the Inertia plugin
- Root Blade template uses Inertia v3 components and data-inertia on `<title>`


taylorotwell:
Hey there! We're no longer updating Breeze in this way. If you want, you could fork it and make it your own!

PR #483, opened 2026-04-07
[2.x] Fix tests running in CI

jackbayliss:
Tests have been broken for some time https://github.com/laravel/breeze/actions/workflows/tests.yml

Noticed this in the installer repo, so checked all of em to ensure we can sleep easy.

I know breeze isnt gonna be majorly updated, but nice to know nothing has broken I guess.

Updated the checkout action while I was here

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,7 +5,7 @@ on:
     branches:
       - master
       - develop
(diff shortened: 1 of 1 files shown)


jackbayliss:
All the failures are a bit much, so ima close it and let it be- for now.

PR #484, opened 2026-05-09
Fix Blade scaffold build failure from missing bootstrap module

azwri:
## Summary
This fixes a Blade stack scaffolding regression where the generated frontend entry imports ./bootstrap but the file is not created, causing Vite to fail during the post-install build.

## Problem
After running Breeze Blade install, the generated app can fail on npm run build with unresolved import for ./bootstrap from resources/js/app.js.

## Root cause
- Blade scaffold writes resources/js/app.js with an import './bootstrap'
- Blade stubs did not provide/copy resources/js/bootstrap.js
- Blade install path did not ensure axios dependency for the bootstrap file

## Changes
- Added Blade stub resources/js/bootstrap.js
- Updated Blade installer to copy that stub into the target app
- Ensured axios is added as a dependency in the Blade install path

## Impact
Fixes first-run Blade scaffold build reliability for Blade installations.

### DIFF
--- a/src/Console/InstallsBladeStack.php
+++ b/src/Console/InstallsBladeStack.php
@@ -25,6 +25,13 @@ trait InstallsBladeStack
             ] + $packages;
         });
 
+        // NPM Dependencies...
+        $this->updateNodePackages(function ($packages) {
+            return [
+                'axios' => '^1.7.4',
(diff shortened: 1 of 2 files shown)


taylorotwell:

